PR #28, opened 2026-07-23
Charts: fix negative-value bars (anchor to zero, not the range floor)

nyblnet:
## Problem
Bars with negative data render wrong. They were anchored to the axis
**range floor** (`baseOf = yOf(range.lo)`) instead of the **zero line**, so
with any negatives:

- positive bars started at the chart floor and stretched up *through* zero (far too tall)
- a bar equal to the range minimum had zero height (invisible)
- negative bars sat on the floor instead of hanging down from zero

(The line *stroke* was already correct — `yOf` handles negatives — but the
line **area fill** shared the same floor baseline, so it was wrong too.)

## Fix
Make `baseOf` return the **zero line**, clamped into the axis range
(`yOf(min(max(0, lo), hi))`), and grow each bar from there toward its value —
up for positive, down for negative. The area fill uses the same baseline, so
areas now fill to zero as well.

All-positive charts are unaffected: with `lo = 0`, `baseOf` still resolves to
the chart floor exactly as before.

## Verification
Repro deck with `bar: [50, -40, 30, -20, 60]` + `line: [20, -25, 45, -10, 35]`:
- **Before:** every bar anchored to the −40 floor; −40 bar invisible; positives crossed zero.
- **After:** +bars grow up from zero, −bars hang down from zero, line crosses zero correctly.

Type-check clean. Independent of the open Slideshow-runner PR (#27).

### DIFF
--- a/slides/src/charts.ts
+++ b/slides/src/charts.ts
@@ -296,7 +296,12 @@ function renderCartesian(svg: SVGSVGElement, d: Digest, sweep: number, view: Vie
     const r = ranges[a] ?? r0
     return G.y + G.h - ((v - r.lo) / (r.hi - r.lo || 1)) * G.h
   }
-  const baseOf = (a = 0) => yOf((ranges[a] ?? r0).lo, a)
+  // baseline for bars / area fill = the ZERO line, clamped into the axis range
+  // (so bars grow up/down from zero, not from the range floor — the negatives fix)
+  const baseOf = (a = 0) => {
+    const r = ranges[a] ?? r0
+    return yOf(Math.min(Math.max(0, r.lo), r.hi), a)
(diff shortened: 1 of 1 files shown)


nyblnet:
Closing as redundant: PR #21 ("Launch fixes: chart zero-baseline for negatives + mobile-Safari pinch") already fixes the negative-bar rendering with the same zero-anchored `baseOf` + bar-geometry approach — and more completely (it also drops the x-axis line to zero, and bundles the iOS-Safari pinch fix). Verified #21's version renders negatives correctly. The two touch the same lines of charts.ts and would conflict, so #21 is the one to merge. Apologies for the duplicate — should have checked #21 first.

PR #55, opened 2026-07-25
Relay: per-room blob quota, expiry cleanup, and a size-limit boundary fix

nyblnet:
Blob storage shipped without quotas: `MAX_BLOB` capped a *single* upload, but
nothing capped a room's total and nothing ever deleted anything. That's an
unmetered write endpoint behind a trust-on-first-use token — the same
unbounded-bill shape we closed for frames, reopened on a different door.

## Quota
`ROOM_BLOB_CAP` (256 MB) metered in the Durable Object, which is the only
component that knows the room. The blob route already called the DO for auth,
so accounting rides on that same call rather than adding a round trip: `PUT`
passes size + key, the DO reserves or refuses `507`, and the client surfaces
`room-full` as **permanent** so it can't retry into a full room forever.

Dedupe doesn't double-charge — a key already counted is admitted without
re-reserving, matching the content-addressed semantics.

## Expiry
The idle alarm now deletes the room's R2 objects **before** wiping DO state.
The DO is the only thing that knows which objects belong to the room, so
wiping its state first would orphan them in the bucket permanently — and
nobody is ever billed less for storage they can't find. Best-effort: an R2
error still lets the wipe proceed rather than wedging the room.

## A boundary bug the quota test found
The client checked **plaintext** length against `MAX_BLOB` while the relay
checks the **ciphertext** it receives, which is larger by the header and
per-chunk IV + GCM tag. An asset just under the limit passed the client and
was refused `413` by the relay.

Added `encodedSize()` and `maxAssetBytes()`. The true ceiling is **8,388,536
plaintext bytes**, which encodes to exactly `MAX_BLOB` — and one byte more is
correctly rejected on both sides. This is the kind of thing that only shows up
when you push a real payload at a real limit.

## Verified
Under `wrangler dev` with R2: upload within quota works and is readable, a
repeat upload dedupes, and the cap engages after **31 max-size uploads
(~254 MB)** with reason `room-full`.

### DIFF
--- a/server/sync-worker/src/worker.js
+++ b/server/sync-worker/src/worker.js
@@ -54,6 +54,12 @@ const ROOM_BYTE_CAP = 96 * 1024 * 1024
 // memory, so a browser encrypting a 64MB asset in one call would need ~320MB.
 // Chunking is a MEMORY requirement, not just resumability.
 const MAX_BLOB = 8 * 1024 * 1024
+// Total blob bytes one room may hold. Frames have ROOM_BYTE_CAP; without the
+// equivalent here, blob storage is an unmetered write endpoint behind nothing
+// but a trust-on-first-use token — the same unbounded-bill shape we closed for
+// frames. Counted in the DO, which is the only component that knows the room.
+const ROOM_BLOB_CAP = 256 * 1024 * 1024
+const BKEY = (k) => `b:${k}`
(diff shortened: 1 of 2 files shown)


nyblnet:
Superseded by #57, which merges this work in so the whole blob-offload feature lands together — avoiding a window where assets over 64KB stop syncing before the session half exists.

PR #59, opened 2026-07-25
Fix #54: a manually-set morph id never morphed

### LINKED ISSUE #54: Setting manually morph ID does not create morph animation between slides
Hi, great project, love it. Thanks for your work!

**What happened**
Setting Same morph ID on two elements does not trigger morph animation between slides. This is quite terrible if you want to edit element back and forth between slides, it makes adding elements with clean animations 

[bug_morph_id.html](https://github.com/user-attachments/files/30372771/bug_morph_id.html)

on previous slides impossible.

The morph animation using ID doesn't work without duplicating the previous slide.
Setting the morph ID of a previous element directly set the pairing key (impossible to clear) and no morph animation.

**Steps to reproduce**
1. Create an element
2. Copy the element to the next slide + set the same morph ID
3. set both slide on Morph Transition
4. play Slideshow

**Environment**
- App version (About dialog, or the shell's `APP_VERSION`): v1.0.8
- Browser + version: Brave 
- OS: MacOS 26.5.2 (25F84)

nyblnet:
Fixes #54.

## The bug

Setting `morphId` to pair an element with one on another slide did nothing. The morph only ever worked via the duplicate-a-slide idiom (same `id` on both slides) — exactly as the reporter described.

## Cause

`data-flip-id` is `morphId || id`, and present.ts pairs the **DOM** by that key correctly. But `modelById()` keyed the **model** maps by `el.id`, while every lookup into them passes a flip id:

```js
const a = fromModel.get(node.dataset.flipId)
const b = toModel.get(node.dataset.flipId)
if (!a || !b) continue          // <- always taken for a morphId element
```

So an element carrying a `morphId` matched in the DOM, then missed its own model entry and fell straight out of the geometry tween. Nothing animated, no error. Elements using the default key were unaffected — which is precisely why "duplicate the slide" worked and "set a morph id" didn't.

## Fix

Key the model maps by the morph key (renamed `modelByMorphKey`). The panel already rejects a `morphId` colliding with another element's effective key on the same slide, so the map stays 1:1.

The bug *was* key-computation drift — two sites deriving the key independently — so `morphId || id` is now a single exported helper `morphKey()` in model.ts, used by render.ts, present.ts and panels.ts. No inline computation of that expression remains in the codebase.

## Verification

Against the reporter's own deck, loaded into a dev build and driven through anim.ts's manual clock (the preview pane never fires rAF, so wall-clock animation can't be sampled there):

| | shape `a` | shape `b` | shape transform through the transition |
|---|---|---|---|
| **pre-fix** | `true` | **`false`** | stays `(none)` the whole time |
| **post-fix** | `true` | `true` | `translate(-241.2px, 140.9px) scale(0.599)` → `translate(-153.9px, 89.9px) scale(0.744)` → `translate(-12.5px, 7.3px) scale(0.979)` → settles |

In both runs the same-id text element on the same slide morphs normally, isolating the defect to the `morphId` path. The trajectory matches the model geometry (`127.4,260,300x200` → `381.1,111.8,519x346`).

`tsc -b` clean.

## Not fixed here

The reporter also notes the override is hard to **clear**. The documented way (type the element's own id) works in code, but the "Pair with" picker offers no unpair option, so there's no affordance. Left as a separate UX change — it needs a new UI string in all 7 catalogs and doesn't belong in a rendering bug fix.

### DIFF
--- a/slides/src/editor/panels.ts
+++ b/slides/src/editor/panels.ts
@@ -5,7 +5,7 @@
 // into a single undo checkpoint.
 
 import type { Store } from '../store'
-import { MEDIA_EMBED_BUDGET, applyChartPalette, defaultChart, uid, type ChartElement, type LineEnding, type MediaElement, type ShapeElement, type Slide, type SlideElement, type TableElement, type TextElement, type TransitionKind } from '../model'
+import { MEDIA_EMBED_BUDGET, applyChartPalette, defaultChart, morphKey, uid, type ChartElement, type LineEnding, type MediaElement, type ShapeElement, type Slide, type SlideElement, type TableElement, type TextElement, type TransitionKind } from '../model'
 import { isMacOS } from '../screens'
 import { CHART_PRESETS } from '../charts'
 import { FONT_CHOICES, firstFamily, injectFonts } from '../fonts'
@@ -459,7 +459,7 @@ export class PropsPanel {
    */
   private buildMorphProps(el: SlideElement) {
     this.section(t('Morph'))
-    const effective = el.morphId || el.id
+    const effective = morphKey(el)
 
     const warn = document.createElement('p')
     warn.className = 'ed-hint ed-morph-warn'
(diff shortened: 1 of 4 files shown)


niemes:
Hey, that was quick! Thank you.

In my local test to fix the bug a one liner worked for my local file, FIY:

In present.ts:
```javascript
function modelById(doc: BentoDoc, index: number): Map<string, SlideElement> {
  const map = new Map<string, SlideElement>()
  // for (const el of doc.slides[index]?.elements ?? []) map.set(el.id, el)
  for (const el of doc.slides[index]?.elements ?? []) map.set(el.morphId || el.id, el)
  return map
}
```

But I didnt dig much...

Cheers

PR #68, opened 2026-07-25
Math mode

### LINKED ISSUE #54: Setting manually morph ID does not create morph animation between slides

dominik-strutz:
**What & why**

This adds a math element which comes in two modes: equation and note. A note lets you write text with formulas inline between `$…$`, so you don't have to drop into LaTeX for a whole text box. Before this, math could not be rendered nicely in Bento. This proposal adds that, and also lets you use (most of) the power of the morph transitions on formulas: two equations on morph-linked slides animate symbol by symbol, so a term crossing the equals sign visibly travels there instead of the formula crossfading. See [The_Light_Clock.bento.html](https://github.com/user-attachments/files/30378051/The_Light_Clock.bento.html) for an (LLM-generated) example.

Two things I tried to be careful about, since they seemed important for a self-contained-file project:

Viewing and presenting need no engine and no network. The rendered formula is baked into the file as SVG, so a deck with math opens exactly like any other Bento file. MathJax is only fetched while you are authoring, and never bundled — it is bigger than the whole shell.
The engine is hash-pinned and checked before it is allowed to run, from every source including bento.page, and Offline mode blocks it like everything else.
**For now the point of this request is mostly to serve as inspiration and a showcase of what is possible. I (and some people in the issues) would love to be able to use proper math rendering in Bento.**

If you approve of the current way of implementing this, I am happy to work some more on it and make sure it's fully up to the standard of the main branch. If you have a better idea how to do this, or if a plugin-style system for such additions is planned in the future, feel free to take whatever you want from this branch. At least for now I will keep my fork up to date, since this is extremely useful for me at work.

A few small things in here aren't math that were flagged up in the LLM review. You fixed #54 independently while this was open, so I dropped my version in favour of yours, but the other half is still needed: chart data-morph was pairing on bare id, so a morphId-paired chart moved while its series data snapped. There's also a morph fix that isn't math-specific but that the symbol morphing depends on (runMorph pivoted about the top-left while applyElementFrame pivots about the centre, so rotated morphs snapped on the final frame), and a one-line repair to a motion-path string whose key had drifted out of sync in all 7 catalogs. Happy to split any of these out.

Disclaimer: all of the code is generated by Claude (mostly Opus 5), I have no experience with TypeScript.

**How I verified it**

Played around with the math mode features, but did not test extensively by hand.

The mechanical checks do pass: tsc -b and npm run build:single are clean, the new UI strings are in all 7 catalogs, and the example deck above renders and boots with no console errors. The branch is rebased onto main at 1.0.9.

**Checklist**

Skipped for now, since this is still more of a proposal/inspiration than an addition I am 100% confident in. The mechanical parts are done though: changelog entry, all 7 i18n catalogs, no version bump.


nyblnet:
Thanks for the submission. There's real craft here — naming the field `baked` rather than `html` so `crdt.ts` doesn't route it into the character-level RGA is a subtle catch, and the id scoping, engine pinning, i18n coverage and doc updates are all thorough.

**It can't land in this shape.** `docs/PLATFORM.md` §1: a Bento file must work with *"no backend, no CDN, no network required to open, **edit**, present, or save. Anything that adds a runtime network dependency to the core document lifecycle is off-platform."* Math is a first-class toolbar item beside Text and Shape, so authoring a formula is a core lifecycle action — and it fetches ~2 MB of third-party JS and injects it as a `<script>` into the app's own origin, which holds decrypted deck content, the collab room key, and the ECDSA private keys.

The SHA-256 pin doesn't resolve that. `update.ts` won't apply Bento's own code without an ECDSA signature from an offline key; this executes someone else's code on the strength of a hash, which proves the bytes are as expected but not that they're safe, and can't be revoked. Merging it would also demote "no CDN" from rule to preference, and the next proposal would cite it.

The model, renderer, sanitiser and per-symbol morphing don't depend on the fetch — baked math already renders with zero runtime. Two paths I'd consider: an in-house LaTeX subset renderer (the charts-lite precedent — ECharts was 630 KB, 47% of the shell, and was replaced rather than kept), or toolchain-only baking via your existing `scripts/bake-math.mjs`, keeping the app pure and moving the dependency to the authoring toolchain as embedded fonts already do.

Don't rework anything yet — whether to record an explicit exception is my call, and I haven't made it.

Two issues that apply regardless:

- `bento.page/vendor/tex-svg.js` is 404 and no release script publishes it, so the preferred source never wins and every author falls through to jsDelivr.
- `safeMathStyle` returns `true` for any SVG-namespaced element, so a crafted `baked` value on the root `<svg>` could carry `position:fixed;inset:0;…` and overlay the editor. `sanitizeHtml` strips every attribute from text elements; this should have a property allowlist.

Also conflicts with `main`, and CI needs maintainer approval per run until a first PR lands.

nyblnet:
Closing this for now as I'm rolling out Temml in the next release.

PR #78, opened 2026-07-25
bento/tray: a native iOS host that runs any self-contained HTML document

nyblnet:
Every browser on iOS is WebKit, so the File System Access API doesn't exist there — no in-place save, no autosave write-back, no in-place self-update. `UIDocument` is the only way to write back to the user's real file, and only a native app can use it.

## The design decision

**The app supplies file access and nothing else.** It bundles no runtime for rendering and has no opinion about which Bento version a deck carries.

The deck runs its **own embedded runtime**, exactly as in Safari, so it self-updates through the normal signed channel — iOS users get the same release the same day, with **no App Store submission per release** and no second release train.

Bundling a shell and rendering every deck with it (my first proposal) would have parked iOS behind a review queue forever and let the bundled copy drift. It isn't needed: 2.5.2 concerns downloading code that changes features **of the app**, and this app behaves identically whatever a document contains. What updates is user content — the same as any page a browser renders.

## No changes to Bento

`save.ts` tests exactly `typeof window.showSaveFilePicker === 'function'` and needs three methods, so `bridge.js` polyfills that over UIDocument. Two consequences:

- Every in-place path (⌘S, autosave, self-update, the capability-aware messaging from #74) already routes through that one function.
- **Every deck ever saved works**, including files whose runtime predates this app. A bespoke `__bentoHost` bridge would only have helped decks re-saved afterwards — i.e. none of the existing ones.

## Which file a save targets — deterministic, not heuristic

Bento only reaches a picker when it holds no handle; afterwards ⌘S, autosave and update all reuse it. So the **first** request is the open document (no UI) and any **later** one is a genuine Save-As or export.

An earlier draft compared `suggestedName` to the open filename. Testing showed that prompts on **every save**, because Bento derives that name from the deck *title*, which rarely matches. Worth knowing before anyone reintroduces it.

## Two details that carry weight
- The deck is served through a **custom scheme**, not `loadFileURL`. A `file://` page in WKWebView has an opaque, unstable origin, which would silently break the autosave backstop, the per-device collab member key, and locale/motion prefs — and would make relay fetches arrive as `Origin: null`.
- `bridge.js` is injected **`.atDocumentStart`**, because capability is read during boot.

## Verified
The save contract against the **real** Bento build in a browser, with the native side emulated over the same protocol:

| | |
|---|---|
| ⌘S | writes the open document, no prompt — 899KB, `#bento-doc` intact, no stray script-close |
| autosave write-back | reuses the handle, writes silently, tag reads "Saved" |
| "Save a copy…" | prompts for a destination, open document untouched |

## NOT verified
**None of the Swift has been compiled or run.** This machine has Command Line Tools but no Xcode, so there's no `xcodebuild` and no Simulator; `swiftc -parse` checks syntax only and can't resolve UIKit or WebKit. **Treat every Swift file as unreviewed until it builds.**

Also outstanding: `Resources/starter.bento.html` (the seed for new documents) isn't wired up, so document creation currently fails; plus icon, launch screen, signing, and an Apple Developer account.

Merging this is a scaffold-landing, not a shipping decision.


nyblnet:
Superseded by #85 — same commits, recreated from the renamed `tray` branch. GitHub can't retarget a PR's head branch, so this had to be reopened rather than moved.

PR #81, opened 2026-07-25
i18n: language pack pipeline, with Korean as the first pack

nyblnet:
Builds the machinery from `docs/i18n-packs.md` (#77) so adding a language stops costing shell bytes. **Korean is the first pack and the proof it works.**

Stacked: this → #79 (Portuguese) → #75 (key-once packing).

## Why pipeline before content

Every i18n change in this series was verified the same way: call `t()` through the **real lookup path** and confirm it returns a translation rather than the English fallback. Without a loader that check is impossible — generating catalogs first would mean thousands of machine-drafted strings with **zero evidence any of them resolve**, and a mistyped key falls back to English silently, forever.

## Kernel — `addPack()`

Layers a pack over the bundled core, keyed by language.

- **Packs are consulted before the core**, so a pack can also *correct* a bundled language without cutting an app release — which is what turns native review into a continuous process instead of a release blocker.
- **Deliberately tolerant.** A pack built against an older app simply lacks the newer keys; each miss falls through to the core and then to English. A stale pack degrades string by string rather than failing to load.
- **Rejects a pack built for a different app** — its keys would be *wrong*, not merely missing, which is the one case worth refusing outright.
- `registerI18n` re-appends pack languages to the picker: block order in the shell isn't guaranteed, and a pack loaded *before* registration would otherwise lose its entry.

## Build — `build-i18n.mjs --packs <dir>`

Emits one JSON per language from `slides/src/i18n/packs/*.ts`. Those are authored exactly like a core catalog, but **nothing imports that directory**, so it never enters the module graph. Errors if a language is listed as both bundled and packed.

## Cost

| | |
|---|---|
| pipeline, added to shell | **+200 B** |
| Korean pack | **7,207 B — entirely outside the shell** |

## Verification (end-to-end, in the browser)

| | |
|---|---|
| before loading | `ko` absent from picker; `t('Delete slide')` → English |
| after `addPack()` | `ko` in picker labelled 한국어; `슬라이드 삭제`, `저장`, `공유`; interpolation intact (`슬라이드 4`) |
| **degradation** | `Version history…` — a key this partial pack does **not** carry — correctly returns English, while carried keys resolve in Korean |
| wrong app | a pack marked `app:'spaces'` is rejected |

Korean is absent from `packed.ts` (not bundled). Core `--check` current, splice conformance gate passes, app + kernel typecheck clean.

## Korean is partial — deliberately

**262 of 650 strings**, covering the most visible UI. That's a *valid* pack by design: the rest fall back per string, which is exactly the degradation demonstrated above. Completing it is the obvious next step.

**Issue #17 is a volunteer offering to write this catalog.** I'd suggest replying to offer them review and ownership of the Korean pack rather than presenting it as done — a reviewed pack is worth considerably more than a machine-drafted one, and the pack format means they can revise it without waiting on a release.

## Still unbuilt from the spec

Signing, the manifest `packs` array, in-app "Add language…", **update-carries-packs**, and `shell-gate` coverage for a pack-carrying shell. Packs currently load via `window.bento.i18n.addPack()` — enough to prove the mechanism, not yet a user-facing feature.


nyblnet:
**Korean is now complete — all 650 strings** (was 262).

There was no technical reason to stop at a partial catalog; that was a stopping point, not a design decision. Corrected.

| | |
|---|---|
| pack size | **58,137 B** |
| contribution to the shell | **0 B** — `ko` remains absent from `packed.ts` |

**Verified:** all 650 resolve through the real lookup path — zero fell back to English, zero mismatched against the pack source. `ko` shows in the picker as 한국어; interpolation (`버전 1.0.11을(를) 사용할 수 있습니다.`) and inline HTML survive. Core `--check` current, splice conformance gate passes, `tsc -b` clean.

The four escape-sensitive strings were generated from the canonical catalog rather than hand-typed — two *pairs* of keys differ only in backslash/newline escaping, and a mistyped escape yields a key that silently never matches. The assembler now also fails on any key that is missing **or** extra, so a pack can't drift from the catalog it targets.

Still machine-drafted. The offer to #17's volunteer stands and is worth making: a pack can be corrected and re-released without an app release, so their review isn't blocked on anything in this PR.

PR #84, opened 2026-07-25
Starter deck: demonstrate symbol morph with a rearranging formula

nyblnet:
Symbol-level maths morph is **invisible unless you see it move** — a screenshot can't show it, and the starter deck is the feature tour. So it gets one beat.

Stacked on #83 → #76.

## Deliberately small

The formula sits quietly under the caption on the existing **"Morph."** slide, then becomes the point of **one** new slide where it rearranges:

```
$a + b = c$   →   $a = c - b$
```

Same element id on both, so `b` is *seen* to travel across the equals sign while the four `sd-tile-*` shapes continue their own morph.

This extends the morph story rather than opening a separate "maths" topic. The tour is shared by everyone, and a slide teaching LaTeX would spend attention on behalf of users who don't need it — depth belongs in a gallery deck. **+564 B** in every shipped shell.

## The caption escapes its dollars on purpose

It writes `\$…\$`. Writing them bare renders them *as maths* — which is exactly what happened on my first attempt, turning "plain `$…$` in a text box" into "plain … in a text box", with the ellipsis silently converted into a formula.

It's the first thing anyone copying this deck will hit, so the deck should demonstrate the escape rather than trip over it.

## Verification

On the real deck transition, driving anim's manual clock:

| symbol | trajectory |
|---|---|
| **`b`** | **−250.4px → −11.0px → settles** |
| `c` | −52.7px → −2.3px → settles |
| `=` | +2.6px → +0.1px → settles |
| `a` | +5.5px → +0.2px → settles |

`b`'s horizontal travel is an **order of magnitude** larger than any other symbol — the signature of a term crossing the equals sign. Caption renders literal `$…$` with no `<math>` inside it. Splice conformance gate passes, `tsc -b` clean.

⚠️ **A note on screenshots of this branch:** the preview pane never fires `requestAnimationFrame`, so morphs freeze mid-flight and a naive screenshot shows a half-transformed slide that looks broken. It isn't — driving the clock to completion settles it. Worth knowing before reviewing visually.

## Merge order

**Must not land before #76.** The starter deck ships inside every build, so maths content without the renderer would make every fresh Bento open showing a literal `$a + b = c$` as its fourth slide.

### DIFF
--- a/slides/src/starterdeck.ts
+++ b/slides/src/starterdeck.ts
@@ -55,6 +55,8 @@ const T_A = 'sd-tile-a' // amber
 const T_B = 'sd-tile-b' // blue
 const T_C = 'sd-tile-c' // paper/white
 const T_D = 'sd-tile-d' // ink panel / card
+/** The formula that rearranges across the morph beat — one id, two slides. */
+const EQ = 'sd-eq'
 const TITLE = 'sd-title'
 const KICKER = 'sd-kicker'
 const GLOW = 'sd-glow'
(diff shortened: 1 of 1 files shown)


nyblnet:
**Upgraded the example — the quadratic formula now, not `a + b = c`.**

```
$ax^2 + bx + c = 0$   →   $$x = \frac{-b \pm \sqrt{b^2-4ac}}{2a}$$
```

The old pair proved the mechanism but showed nothing of the typesetting. This one exercises the whole renderer at a glance — **fraction, radical, superscript, ±** — in the one equation everyone recognises. Display mode (`$$`) on the hero so the fraction and radical set at full size.

It's also a far better morph. Measured 18% into the transition:

| symbol | travel |
|---|---|
| **a** | **573px** |
| **c** | **430px** |
| 2 | 361px |
| = | 137px |
| b | 89px |
| x | 12px (mostly vertical) |

All settled by 68%. The previous example moved one symbol 250px; this moves four symbols several hundred each — `a` and `c` fly out of the quadratic and into the discriminant and the denominator. That's the difference between "you can see it worked" and "that's worth showing someone".

Speaker notes rewritten to match — they described `b` crossing an equals sign that no longer exists — and now carry the point worth making: the document stores the LaTeX, not a picture, not a font, with nothing fetched at runtime.

**+780 B** over the symbol-morph branch. Gate passes, `tsc -b` clean. Verified `mfrac`/`msqrt`/`msup`/`display=block` all present, the before-state still renders as maths on the previous slide, and the caption still shows a literal escaped `$…$` with no `<math>` in it.

PR #88, opened 2026-07-26
feat(slides): export editable PowerPoint files

bluegitter:
## Summary
- add an editable PPTX export path for bento/slides decks
- map common slide primitives, images, media placeholders, tables, charts, links, and speaker notes into PowerPoint output
- document current export coverage and include PptxGenJS in bundled notices

## Commit structure
- fix(build): extract stylesheet outside bundled scripts
- feat(slides): export editable PowerPoint files

## Validation
- npx -y node@24 scripts/build-i18n.mjs --check
- cd slides && node_modules/.bin/tsc -b
- cd slides && node_modules/.bin/tsc -p ../kernel
- cd slides && npm run build:single
- node scripts/shell-gate.mjs slides/dist-single/Bento_Slides.bento.html


nyblnet:
Thanks for submitting this! Powerpoint export is one of the most asked features but I'm still trying to figure out the best way to handle it without making the runtime significantly larger. I'm also worried that people will use Bento as a pptx creator which shouldn't be the case. A lot of the nice visual elements will be los in Pptx format, but also one pptx deck shared is one less Bento deck shared. I'll try to find a middle way here and will use your PR as a base.

thiswillbeyourgithub:
Hi,

Just to add my 2 cents on why I think this feature is important: the only reason I haven't taken the jump to bento so far is that I'd feel safer knowing I can import/export from/to powerpoint. Slideshows are for me only something I do for relatively high stakes things, and only professionally. And turns out that I don't know in advance if someone will ask me for my slides for example and i'd prefer not giving html file to people with limited computer knowledge.


IIUC right now my only bet would be to use the claude skills and tell it to convert to a .pptx, right?

PR #97, opened 2026-07-26
Implement Ghost Cut support for slide elements

### LINKED ISSUE #56: Ghost
**The problem / use case**
Moving text using traditional Cut & Paste is flawed:

* **Overwrites Clipboard:** Cutting replaces existing clipboard contents; undoing restores the document text but loses the previous clipboard data.

* **Document Reflow:** Cutting immediately shifts text layout, making it harder to maintain visual context and relocate the target insertion point.

Image

* **Non-Atomic Operations:** Moving text requires separate cut and paste steps, making single-step undo impossible—especially if intermediate edits occur.

**What you'd like**
Implement "Ghost Cut" functionality for moving text/images/objects:

* **Faded/Inert State (Ctrl+X / Cmd+X):** Selected text fades and becomes non-interactable (uncopyable, unclickable, cursor skips over it) while remaining visible in place without altering the clipboard.

* **Cancel (Esc):** Restores the ghosted text to an active, editable state.
Atomic Move (Ctrl+V / Cmd+V): Removes the ghosted text/images/objects from its origin and inserts it at the current cursor position as a single, fully reversible move operation.

See: https://ishmael.textualize.io/blog/ghost-cut/

senthan-07:
## What & why

This PR implements Ghost Cut for canvas/slide elements in Bento Slides.

Instead of immediately removing selected elements on **Ctrl/Cmd+X**, the editor now creates an editor-only ghost state without modifying the system clipboard. Ghosted elements remain visible but become inert, allowing users to maintain visual context while choosing a new location.

- **Ctrl/Cmd+X** ghosts the selected element(s)
- **Esc** cancels the ghost state and restores the original elements
- **Ctrl/Cmd+V** performs an atomic move by removing the original elements and inserting them at the destination
- The entire move can be undone in a single undo operation

This implementation covers **canvas/slide elements** (such as text boxes, images, shapes, charts, and other slide objects). **Ghost Cut for inline text editing inside an actively edited text box is not included in this PR.**

Additionally, this PR fixes Ghost Cut fading to work correctly with the renderer's inline opacity. Ghost opacity is now derived from each element's model opacity (`opacity * 0.35`) instead of relying on a CSS opacity rule that could be overridden by inline styles.

Closes #56

## How I verified it

- Built successfully using `npm run build:single`
- Manually tested in a browser using mouse and keyboard:
  - Ctrl/Cmd+X ghosts selected element(s)
  - Ghosted elements are faded and cannot be interacted with
  - Esc restores the original elements
  - Ctrl/Cmd+V moves the ghosted element(s)
  - Undo restores the entire move in a single operation
  - Existing element opacity is preserved after cancelling Ghost Cut

## Checklist

- [x] Read the relevant contribution guidelines
- [x] `npm run build:single` succeeds
- [x] No changes to `slides/src/sync` (CRDT tests not required)
- [x] No new UI strings were added
- [x] No document format changes
- [x] No version or release changes


senthan-07:
Demo of the Ghost Cut implementation in this PR.

This video demonstrates:
- Ctrl/Cmd+X creating the ghost state
- Ghosted elements remaining visible but inert
- Esc cancelling the ghost state
- Ctrl/Cmd+V performing the atomic move
- Ctrl/Cmd+Z undoing the move in a single operation
[Ghost_cut.webm](https://github.com/user-attachments/assets/fb3519e4-fc9d-4a2a-865e-4293aef3d884)

PR #98, opened 2026-07-26
Editor: make the topbar fit a phone

nyblnet:
Measured, not guessed: at a **402px viewport the topbar needed 681px**. Label collapsing had already fired — every button icon-only at 36px — so CLAUDE.md's "responsive by hiding text, never scrolling" was only half true. 17 icon buttons plus gaps is an irreducible ~684px floor.

Worse, **nothing clipped it**. `.ed-root` had no `overflow`, so bar overflow became *document* scroll — swiping the toolbar dragged the canvas off screen, and Save sat past the right edge unreachable.

## Changes
- `.ed-root { overflow: hidden }` — the column can never scroll sideways again.
- Under 700px (the threshold the JS already used to collapse panels), `applyPhoneChrome` **reparents** low-priority controls into **＋** and **⋯** dropdowns. Reparenting, never rebuilding: a phone-only copy would need its own listeners and would desync from live state — the dirty dot lives *on* the save button, the comment button carries an armed class.
- Side panels become **overlay drawers**. As flex siblings the 188px sidebar left a 214px canvas: the slide became the smallest thing on screen.
- 44px touch targets; resizer chevrons give way to real ☰ / Format buttons.

**Result at 402px: topbar 681 → 402, page no longer scrolls sideways.**

## Two bugs found while building it
- Icon-only buttons (undo/redo/PDF) have no `<span>`, so the existing rule restoring labels inside menus had nothing to reveal — they'd sit in ⋯ as mystery glyphs. They now borrow their tooltip minus its shortcut ("Redo (⇧⌘Z)" → "Redo"): no new strings, no desktop change.
- The `MediaQueryList` is held on `this`. An unreferenced MQL can be collected along with its listener, leaving the bar folded forever once a window grew.

## Verified
- **402px**: folds, fits exactly, no page scroll, both menus carry the real controls.
- **1280 fresh boot**: insert group intact, undo+redo both present, phone menus empty, phone-only controls hidden, no borrowed labels.

## NOT verified
The live transition across the breakpoint. This harness's viewport resize dispatches **neither** `matchMedia` change **nor** `window.resize` — measured, zero events for both — so crossing 700px can't be exercised here. Both listeners are wired (resize as well as matchMedia, since a phone rotating is exactly that path), but the crossing itself needs a real window drag.

## Known gap
The save-as menu (copy / new deck / template / password) doesn't fit on a phone and is unreachable there for now; it should move into ⋯.

4 new strings across all seven catalogs. Splice gate OK, rig `SEEDS=200` ALL PASS.


nyblnet:
Superseded by #136, which is these same five commits replayed onto `main`.

Rebasing was only part of it — this branch also carried three generated files that shouldn't be in git:

```
tray/ios/BentoTray.xcodeproj/project.pbxproj
tray/ios/BentoTray.xcodeproj/project.xcworkspace/contents.xcworkspacedata
tray/ios/Resources/starter.bento.html   (587KB, regenerated every build)
```

Not a mistake in this PR. The `.gitignore` rules named `tray/BentoTray/`; the tray session renamed that directory to `tray/ios/`, which silently unignored both paths, and a broad `git add` here picked them up. #136 drops them and fixes the rules so it can't recur — those rules had only ever been corrected in an uncommitted working tree, so nothing on `main` was protecting against it.

Also flagged there: this branch switched `.ed-corner-br` from the logical `inset-inline-end` to the physical `right`, which puts the present/zoom cluster in the wrong corner under RTL now that ar/fa/he/ur ship as packs. #136 keeps the logical property with the same `max()` values.

Verified at 402×874: no horizontal overflow, Save on-screen.

PR #123, opened 2026-07-26
Fix chart label and legend style rendering

YishenTu:
**What & why**

charts-lite previously ignored several visual options already present in chart JSON. This made larger template charts difficult to read and caused CJK legend labels to overlap.

This change:
- honors legend font size and weight, marker dimensions, item gaps, and numeric top/bottom placement;
- honors x/y axis-label font size, weight, and color;
- honors axis and split-line widths plus line-series symbol size;
- estimates CJK legend text as full-width and centers visible legend content;
- preserves the legacy safe inset for bottom legends and reserves the correct pie-chart area for top or bottom legends.

The document format remains unchanged and unsupported option keys still degrade gracefully.

**How I verified it**

- `cd slides && node_modules/.bin/tsc -b`
- `cd slides && npm run build:single`
- Browser QA with a combo chart using a 22px top legend, 20px x-axis labels, heavier lines, and larger symbols
- Browser regression QA with the built-in starter deck bottom legend
- Independent adversarial review followed by a clean final re-review

The single-file build succeeds. It still reports the existing CSS minifier warning around `.ed-lang-warn`; this change does not touch that CSS.

**Checklist**
- [x] CRDT convergence rig not required; no `slides/src/sync/` files changed
- [x] No new UI strings
- [x] No document format change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ pre-1.0.
 
 ## [Unreleased]
 
+- **Chart labels and legends now honor their visual options.** The lightweight
+  chart renderer applies configured font sizes and weights to axis labels and
+  legends, respects legend spacing and placement, and measures CJK legend text
+  correctly so localized series names no longer overlap.
+
 - **Updating a file now suggests that file's own name.** When an update asks
   where to save, the dialog is pre-filled with the name of the deck you have
   open rather than one derived from its title — so a file called
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -225,6 +225,12 @@ names provisional.
   `yAxisIndex` (0/1). renderCartesian computes a range per axis, shares gridline
   rows (2nd axis labels on the right, its own nice scale via `fixedTicks`), and
   honours per-axis `min`/`max` + `axisLabel.formatter` ('{value}%'). **Visual
+  text styling**: charts-lite honours `legend.textStyle.fontSize/fontWeight`,
+  legend marker size/gap and numeric `top`/`bottom`, x/y
+  `axisLabel.fontSize/fontWeight`, axis and split-line widths, and line-series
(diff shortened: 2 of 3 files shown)


nyblnet:
Merged — thanks. The CJK legend overlap was a real bug (I reproduced it on `main`: consecutive labels 1.1px apart with markers landing *inside* the previous label's box), and this fixes it cleanly. The three render paths still agree byte-for-byte, dual y-axis and palette inheritance are unaffected, and hostile option shapes degrade without producing `NaN`.

Two follow-ups I'm picking up rather than sending back, so this doesn't sit waiting on you:

**1. Cartesian grid doesn't reserve the legend's height.** `legendH` is computed and applied to `pieTop`/`pieBottom`, but the cartesian grid still uses the fixed `num(g.bottom, 44)` — so a large legend grows into the x-axis labels. Measured on a bar chart with no explicit `grid` and `legend.textStyle.fontSize`: 2 collisions at 20px, 4 at 28px. A chart with a generous explicit `grid` hides it, which is probably why it didn't show up in your testing.

If you'd rather finish this yourself, say so and it's yours — otherwise I'll pick it up in a day or two.

**2. Our starter deck loses contrast — not your bug.** It declares `yAxis[].axisLabel.color: 'rgba(182,193,210,0.5)'`, which the old engine ignored in favour of the x-axis colour. Your change is *more* correct; it just exposed that the deck was tuned against the buggy renderer, so those labels now drop to ~3.2:1 on the panel (below AA). Fixing it in `starterdeck.ts` where it belongs.

One small thing for later: `estimatedTextWidth` treats every codepoint above `0x2ff` as full-width, which sweeps in Cyrillic, Greek, Hebrew, Arabic and Thai — "Продажи" at 16px estimates 112px against an actual 70px, nudging a centred legend off-centre. Narrowing it to the East Asian Wide/Fullwidth ranges would be a nice follow-up if you fancy it.

Issue #194, opened 2026-08-02
Agents cannot measure text: layout containers, auto-height, and window.bento.validate()


nyblnet:
Split out of #160 (finding 7), reported by @thinkbig1979. The other eight
findings there were documentation gaps or silent bugs and are fixed in #193.
This one is a feature, and it is the most substantial thing in that report.

## The problem, stated narrowly

It is not that Bento uses absolute pixel positioning. Absolute pixels are what
let the morph engine, the editor's drag handles and the renderer all work from
one representation, and that is a good trade — a layout engine in the middle
would mean the model no longer says where anything is, which is the property
morph depends on.

The problem is narrower: **an agent cannot measure text.** The rendered height
of a string, for a given font, size, weight and `lineHeight`, is unknowable
without a render. So every card, caption and two-line heading is a guess, and
overflow and collisions are downstream of that one gap.

In the authoring run behind #160, laying out a three-photo row meant computing
`3*288 + 2*112 = 1088` by hand, then repeating equivalent arithmetic for a 2×2
card grid and a three-stat row. That was the largest single consumer of
authoring effort after sourcing images.

This matters more over time, not less. The document is the interchange unit for
AI authoring, and `agents.md` is a supported entry point — so the ergonomics of
generating a deck blind are a first-class concern, not a nicety.

## Prevention

**a. Layout containers resolved at author time.** An element carries something
like `layout: {type: "row"|"column"|"grid", gap, padding, cols}` and its
children supply content plus an optional weight. The document resolves it to
concrete geometry when it loads and writes the pixels back.

The stored file therefore stays absolute-pixel: morph, the editor and the
renderer are untouched, and the declarative form is a pure authoring
convenience that disappears on first save. That is what makes this fit the
codebase rather than fight it. Every grid in the deck from #160 would collapse
to one declaration.

Open questions: whether the resolved geometry is written back into the model
(simple, but the `layout` key then describes something already true) or kept as
a live container the editor re-resolves on edit (more useful, much larger
change, and it puts a layout engine back in the middle). Format additivity
means old shells must ignore `layout` and still render the deck correctly —
which the write-back approach gets for free and the live-container approach
does not.

**b. Auto-height text.** `h: "auto"` on a text element, resolved to a concrete
number at render and stored back. This removes the guess for the element type
that causes nearly all of it, and is much smaller than (a).

**c. A positioning table in the docs.** Done in #193 — `agents.md` now carries
[...]

PR #196, opened 2026-08-02
Let fx.enter run for elements new to a morph slide

nyblnet:
> **Stacked on #195** — base is `validate-api`, because the validator check
> changes with the behaviour. Merge #195 first and this retargets to `main`.

Setting an element to sweep in from the right on a morph slide gave you a small
upward nudge instead.

Not because entrances were dead there — that was my first reading and it was
wrong. `runMorph` already gives every unpartnered incoming element an entrance
of its own. The problem is that the entrance is **fixed**: `fx.enter`'s
direction, `enterDur` and `order` were all discarded in favour of a 14px rise.

## The split

The same one the count-up fix used, applied per element:

| on a morph arrival | behaviour |
|---|---|
| **has a morph partner** | travelling from its previous-slide frame, so an entrance would fight the morph — `fx.enter` stays skipped |
| **no partner** | new to this slide, nothing to fight, and the author named a direction — it runs, honouring `enterDur` and staggering by `fx.order` |
| **no partner, no `fx.enter`** | keeps the automatic fade-and-rise |

## Blast radius

Exactly the decks that already wrote `fx.enter` on a morph arrival — where it
currently does nothing. **A deck that never asked for an entrance cannot
change.**

Measured in present mode against the manual animation clock:

| element | starts at | |
|---|---|---|
| `carried` — has a partner | opacity 1, x −36 | morphs from its old position, entrance skipped |
| `plain` — new, no `fx` | opacity 0, y 14 | **unchanged** — the regression control |
| `left` — new, `slide-left` | opacity 0, **x 120** | sweeps in from the right (was y 14) |
| `down` — new, `slide-down` | opacity 0, **y −120** | drops in |

## Also

The direction table now lives in one place (`enterSpec`) rather than being
duplicated across the two entrance paths, so a new direction cannot work on
ordinary slides and quietly do nothing on morph arrivals.

`validate()`'s check narrows to match — only a *partnered* element loses its
entrance — and the rig asserts both directions: a new element is not reported,
a morphing one is. 21 checks.

Closes the last of the entrance half of #194's neighbourhood; `h: "auto"` and
layout containers are still open there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ pre-1.0.
 
 ## [Unreleased]
 
+- **Entrance animations work on morph slides.** Setting an element to sweep in
+  from the right on a morph slide gave you a small upward nudge instead: the
+  morph supplied one fixed entrance for everything new on the slide, and
+  `fx.enter`'s direction, duration and order were discarded. An element that is
+  new to a morph slide now enters the way you asked. One that morphs in from
+  the previous slide still ignores `fx.enter` — it is already in motion, and an
+  entrance would fight the tween — and elements with no `fx.enter` keep the
+  automatic fade-and-rise, so nothing changes in a deck that did not ask for it.
+
 - **`window.bento.validate()` — see what the runtime silently swallows.**
   Almost everything that goes wrong in a generated deck fails quietly: a typo'd
   property is ignored, a `dash-march` loop on a solid stroke animates nothing,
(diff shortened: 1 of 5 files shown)


nyblnet:
Superseded by #197 — same commit rebased onto `main`. GitHub closed this one automatically when `validate-api` was deleted on merging #195, and a PR whose base branch no longer exists can't be reopened.

PR #200, opened 2026-08-02
Pan the canvas by dragging, and past the slide's edges

### LINKED ISSUE #166: Panning with middle click
### Discussed in https://github.com/nyblnet/bento/discussions/144

<div type='discussions-op-text'>

<sup>Originally posted by **gcgbarbosa** July 27, 2026</sup>
Right now, the only way to pan is through the scrollbars.
Also panning does not go beyond the edges of the slides, which limits visibility.

It would be nice to have those two features implemented.</div>

nyblnet:
Closes #166 — both halves of it.

## Middle-drag pans

The scrollbars were the only way to move a zoomed slide, which puts the control
at the edge of the screen while the work is in the middle of it. Middle-drag is
what design tools do, and unlike space-drag it cannot collide with typing a
space.

Bound in the **capture phase**, because Selecto is on the same scroller and
would otherwise read the press as the start of a marquee. Move/up listeners go
on `window` so a pan survives the pointer leaving the canvas, and `auxclick` is
suppressed so X11 doesn't paste the selection on release.

Measured: a 150×90 drag scrolls exactly 150×90, the cursor becomes `grabbing`
and restores, and movement after release does not pan.

## Pan room past the edges

Scrolling stopped dead at the slide's edges, so at high zoom a corner element
could never be moved off the corner of the screen to be worked on. There is now
half a viewport of padding once the stage outgrows the window — exactly enough
for any point on the slide to reach the middle.

| | padding | scroll range | pannable |
|---|---|---|---|
| at fit (83%) | `0px` | 1240 × 673 = viewport | no |
| zoomed (161%) | `329px 620px` | 3303 × 1818 | yes |
| back at fit | `0px` | 1240 × 673 | no |

At 161% scrolled hard to the origin, the slide's left edge sits at 635px in a
1270px viewport — centred, to within rounding.

**None at all while the slide fits**, and that is not just tidiness about
scrollbars. The plain-wheel slide-nav stands down whenever the canvas is
pannable, so padding at fit zoom would have silently cost the wheel gesture
that walks slides. Checked both ways: at fit, padding is `0`, the scroller is
not pannable, and a wheel still advances the slide.

`clientWidth` is the padding box, so writing padding in `relayout` cannot
disturb the `fitScale` computed from it two lines earlier — no feedback loop.

## Help

A Canvas section in the `?` overlay, because a middle-drag nobody mentions is a
middle-drag nobody finds. Five new strings, translated into all 8 core locales.

Asked for by gcgbarbosa in #144.


nyblnet:
Updated after a UX review: **space-drag is now the primary gesture**, with middle-drag kept as the secondary.

Middle-drag alone was the wrong call, for a reason I should have caught given where this project is developed — **no Apple trackpad or Magic Mouse has a middle button.** The gesture #166 asked for is real and standard, but as the *only* pointer way to pan it is unavailable to a large share of the people who'd want it. Space-drag is the near-universal convention (Figma, Illustrator, Photoshop, Miro, Canva, Excalidraw); none of them ship middle-drag alone.

The objection I'd raised against space — that it collides with typing — is one this code was already equipped to answer. Space is otherwise unbound in the editor, and `canvas.ts` already tracks `editing`/`editingCell`/`isPathEditing` together for exactly this kind of question. `canGrabSpace()` adds the `activeElement` test so panel fields keep their spaces too.

Verified while editing text: the keydown is **not** swallowed, the canvas does **not** arm, and a drag does **not** pan.

One real bug found in testing. A probe that never sent `keyup` left `spaceHeld` stuck true, and the next case panned when it shouldn't have. That's not just a test artefact — a keyup can genuinely be missed (native dialog, focus leaving mid-hold), and a canvas stuck armed would swallow the click that starts a text edit. `canGrabSpace()` is now re-checked at **mousedown**, not only when space arms. Verified: armed → editing begins with no keyup → drag pans by exactly `[0, 0]`.

**Two-finger scroll is deliberately unchanged.** It pans when zoomed past fit and walks slides when the slide fits — the presentation-tool convention (PowerPoint, Slides, Keynote) rather than the canvas-tool one (Figma). It isn't really a mode: when the whole slide fits there is nothing to pan *to*, so navigating is the only useful thing that gesture could do.

Help overlay leads with space-drag; two more strings across all 8 core locales.

PR #221, opened 2026-08-03
changelog: the collab property-removal crash

nyblnet:
The reason for the next release, with no CHANGELOG entry — so `release.mjs` would sign a manifest whose notes never mention it.

Those notes ride in the signed envelope and are what people read in the About dialog while deciding whether to update. For a fix to a crash they have been hitting with no workaround, it is the one line that matters.

Written user-facing rather than mechanically: the symptom is *"my collaborators stopped seeing my changes"*, and the trigger is taking a formatting option away — solid fill, outline off, ungroup, unlink, clear a click target.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,18 +13,18 @@ pre-1.0.
 
 ## [1.0.15] — 2026-08-03
 
-- **Fix: one collaborator's edit could crash everyone else in the room.**
-  Removing a property from an element travels as an op the receiving side
-  could not read, and every other participant's editor threw on arrival. The
-  person making the edit saw nothing wrong at all.
+- **Fix: removing a formatting option no longer disconnects the people you are
+  working with.** While a live session was running, taking something *away* —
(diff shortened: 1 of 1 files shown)


nyblnet:
Rebased onto `main` and resolved so this **replaces** the entry that went in with the release commit (#222), rather than sitting alongside it — we'd written one each for the same fix and the changelog would otherwise carry both.

Yours is the better one and is what's kept: it leads with what the user *did* ("removing a formatting option") and what they *saw* ("disconnects the people you are working with"), where mine led with the mechanism. For someone reading the About dialog mid-decision, the action they were performing is the useful hook.

I kept one sentence from mine — that the convergence rig **now generates removals**. Yours notes the tests only ever added, which explains the gap; it's worth also saying the gap is closed, since that's what stops the next op shape slipping through the same way.

Signed notes now lead with:

```
• Fix: removing a formatting option no longer disconnects the people you are working with.
```

PR #226, opened 2026-08-03
bento/spaces 0.1.0 — the first release

nyblnet:
# bento/spaces — the first release

One HTML file is a whole space: a tree of pages, the reader, and the editor.
No account, no server, no sidecar folder.

## What it does

- **Pages nest**; the sidebar is the tree, and drag reorders both pages and blocks.
- **Writing gets out of the way** — markdown as you type, a `/` menu at the caret,
  and `[[` to link a page by name (offering to create it).
- **Links go both ways.** Backlinks are derived, so a space stays navigable
  with nothing to maintain.
- **⌘K** jumps to any page or block; **⌘F** finds and replaces across the space.
- **Images** are downscaled before embedding and stored once per content hash —
  a 4.9 MB photo embeds as 33 KB — with the original one click away.
- **Reading view**, printing and PDF share the renderer, so they agree.
- **Nine languages**, following the reader rather than the document.
- **Archive rather than delete**, because the file is the only copy there is.
- Password protection, autosave + recovery, signed self-update.

## Four things found by measuring, not reading

Each was silent, each is now gated, and every gate has a negative control.

| | Found | Fixed |
|---|---|---|
| **Script ran on open** | `createElement('div').innerHTML = hostile` FIRES `` — the div is detached, but its elements are in the live document, so resources load. The sanitizer must parse hostile markup before stripping it, so the payload ran **before** the strip, at render time, on merely opening a mailed file. | One inert `DOMParser` parse for every untrusted-html path. The guard found a fourth call site — markdown export, which ran what it was exporting. |
| **Tracking pixel** | A space carrying a remote `` requested it on open: the reader's IP and the moment they read your document, sent to whoever wrote the file. Also breaks PLATFORM §1. | Remote images wait for the reader, behind a placeholder naming the host. Consent is per-url, per-session, never in the file. Costs authors nothing — the editor only ever writes `asset:`. |
| **Password defeated** | `putRecovery` ran unconditionally on a 2.5s debounce, so an **encrypted** space wrote its full plaintext into IndexedDB every few seconds. | Skipped while encrypted; setting a password clears what was already written. Measured in both directions. |
| **Replace-all lied** | The readout counted matching *blocks* ("2 found"), the dialog quoted that as "2 occurrence(s)", and the sweep changed 4. | Counting and replacing share one traversal, so the number shown is the number changed — including *not* counting a match split across markup, which cannot be replaced. |

## Release plumbing

`release.mjs` read the root changelog unconditionally, so a spaces release would
have **signed slides' release notes** into the spaces manifest — shown to every
spaces user, inside an envelope that cannot be re-signed for the same version.
dash hit this independently (#234); both now go through one `changelogPath()`,
shared by the signing path and by the new `--print-notes`, which shows exactly
what would be signed and exits before anything is built or wiped.

Also: each app publishes its agent guide at `bento.page/<app>/agents.md` (the URL
the guides already advertised but nothing wrote), and tags are prefixed per app —
slides keeps bare `vX.Y.Z`, everything else is `<app>-vX.Y.Z`.

## Not in this release

Collaboration. The sync engine is slides-shaped; PLATFORM §10 permits shipping
without collab rather than with a half-secure version. `docs/spaces-agents.md`
says so plainly.


nyblnet:
Two findings from reading this branch closely while deriving `bento/dash`'s M0 as a delta from it. Both are small; the first is the kind that only shows up once.

Everything else here transfers directly — the tagged `ParseResult`, flat block properties, deterministic id repair and the byte-capped undo are all being copied rather than re-derived, and two of them were arrived at independently by dash's design work, which is about as good a signal as a design gets.

---

### 1. The oversized-document case boots the starter over live data

`parseDoc`'s contract is that **only an absent or empty block** reaches the starter, and everything else refuses. But the distinction it depends on is already gone by the time it is called:

```js
// kernel/src/save.ts:42-46
export function readEmbeddedDoc(): string | null {
  const block = document.getElementById(DATA_BLOCK_ID)
  const text = block?.textContent?.trim()
  return text || null      // "no element" and "element with empty text" — same answer
}
```

That matters because of a measured failure mode. At a large enough document the `#bento-doc` element is still **present**, with exactly one child text node of length **zero** — `textContent` returns `''`, no throw. So:

- `readEmbeddedDoc()` → `null`
- `parseDoc('')` → `{ ok: false, err: 'empty' }`
- `main.ts:45` → the one branch that boots `starterDoc()`
- first ⌘S → the user's document is replaced by an empty space, permanently

This is exactly the loss the tagged result was built to prevent, arriving through the one door left open. Spaces can't currently express it, because the kernel flattened the two cases before `parseDoc` sees them.

Cheapest fix is at the boot site rather than inside `parseDoc` — the DOM still has the evidence:

```js
const el = document.getElementById('bento-doc')
const embedded = readEmbeddedDoc()
if (el && embedded === null) { /* present but unreadable — refuse, never starter */ }
```

Dash will need this regardless, so if you'd rather leave it, say so and I'll add the distinction kernel-side with both apps as consumers.

### 2. `putRecovery` is unguarded, so an encrypted space snapshots as plaintext

```js
// spaces/src/main.ts:207
timer = setTimeout(() => { void putRecovery(store.doc) }, 2500)
```

Slides guards the same call:

```js
// slides/src/editor/editor.ts:2062-2066
// Never write an encrypted deck's plaintext to IndexedDB; its file
// write-back below stays encrypted via serializeAuto.
if (!isEncryptionActive()) {
  snapshotted = await putRecovery(doc)
  ...
}
```

And the kernel states the rule in its own header (`kernel/src/autosave.ts:14-16`): *"Encrypted decks are never snapshotted here (that would write plaintext to disk)."*

One-line fix at the call site. But the structural point is that the rule is **documented in the kernel and enforced in each app**, so every new app has to remember it independently — and the second one didn't. Worth moving the guard into `putRecovery`/`addVersion` themselves, where it can't be forgotten. Happy to do that as a separate kernel PR with slides and spaces as the two consumers, if you want it out of this one.

---

Unrelated and non-blocking: `docs/spaces-agents.md` looks orphaned — nothing in the repo references it, no build step writes it into `site/`, and the flat hyphenated name can't produce the `bento.page/<app>/agents.md` URL that `docs/agents.md` advertises.

PR #239, opened 2026-08-04
The topbar fits itself by measuring, not by width breakpoints 

bouzidanas:
Px breakpoints clipped the end of the bar whenever the same buttons needed more room at the same viewport width (browser zoom, OS text scaling, wider translations, the update chip or avatars appearing). fitTopbar() now measures the bar and steps down a tier while it still overflows: ed-bar-compact hides labels, ed-bar-tight drops the wordmark, ed-bar-fold moves buttons into menus via applyPhoneChrome. Driven by a ResizeObserver on the bar plus a MutationObserver for content changes; the 700px media query keeps only the viewport-scoped phone chrome (drawer panels, touch targets).

**What & why**

The editor topbar used to collapse at fixed pixel breakpoints (1200 for labels, 760 for the wordmark, 700 for the phone fold). The problem is that the room the same buttons need at the same viewport width isn't constant. Browser zoom, OS text scaling, wider translations, and live content like the update chip or the avatar strip all change it, and every one of those cases could clip buttons off the end of the bar. CSS can't see any of this because media queries and container queries measure the container, never whether the content actually fits in it.

So the tiers are now classes instead of breakpoints. A new fitTopbar() in editor.ts starts from the widest layout and steps down while the bar still overflows its own box (scrollWidth against clientWidth, plus a check for the title input getting crushed, since it's the only shrinkable thing in the bar and absorbs all the pressure first). ed-bar-compact hides button labels, ed-bar-tight drops the wordmark, and ed-bar-fold moves buttons into menus through the existing applyPhoneChrome reparenting. A ResizeObserver on the bar catches width changes and a MutationObserver catches content changes. The 700px media query stays, but only for the genuinely viewport-scoped phone chrome (overlay drawer panels and 44px touch targets), and phones always fold like before.

A few things worth knowing about how this behaves. The mutation observer watches child changes, text, and the style and hidden attributes, so if future bar content changes its width purely through a class toggle it won't trigger a re-fit until the next resize (easy one-line filter tweak when that comes up). Re-fits are also skipped while a dropdown is open, because re-fitting would slam it shut; in the rare case content grows during that window the bar can sit slightly overflowed until the next event, and then fixes itself. And debugging the bar's state now means looking at which ed-bar classes were applied rather than reading the stylesheet.

**How I verified it**

Verified in the browser at 1500, 1100, 750, and 640px, and with CSS zoom at a fixed viewport (the exact case the old breakpoints couldn't see). tsc, build:single, shell-gate, i18n coverage, and the preview, autosave, validate, clipboard, layouts, and savepurpose rigs all pass.

**Checklist**

- [x] Ran `node scripts/test-sync.ts` if I touched `slides/src/sync/`
- [x] New UI strings added to every catalog in `slides/src/i18n/`
- [x] Document format changes are additive and backward-compatible

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -523,8 +523,12 @@ names provisional.
   all (verified against pre-change code too) — resize behavior needs a
   real mouse. Present: real fullscreen via overlay.requestFullscreen at start +
   F toggle (denied requests degrade to tab-fill — that IS the testing/
-  sharing mode). Topbar is responsive by HIDING TEXT, never scrolling: labels
-  collapse to icons <1200px, the wordmark collapses to the mark <760px.
+  sharing mode). Topbar is responsive by HIDING TEXT, never scrolling — and by
+  MEASURING, not width breakpoints (zoom/OS text scale/locale width made px
+  queries clip the bar): editor.ts fitTopbar() steps down tier classes while
+  the bar overflows (ed-bar-compact hides labels, ed-bar-tight the wordmark,
+  ed-bar-fold folds into menus via applyPhoneChrome), driven by a Resize- +
+  MutationObserver on the bar.
   Panel show/hide lives ON the resizer strips as chevron tabs (docked
   flush to the screen edge when collapsed); phones (<700px) boot with
   both panels collapsed (canvas-first; chevrons/[/] bring them back).
--- a/slides/src/editor/editor.ts
+++ b/slides/src/editor/editor.ts
@@ -424,25 +424,34 @@ export class Editor {
       for (const k of kids) this.phoneChrome.homeOf.set(k, g)
     }
 
-    // drive it now and whenever the query flips
-    // Held on `this` deliberately: a MediaQueryList that nothing references can
-    // be collected along with its listener, and the bar then never unfolds when
-    // the window grows — the CSS flips but the JS half silently stops.
-    this.phoneQuery = window.matchMedia('(max-width: 700px)')
+    // drive it now and whenever the bar's size or content changes.
     // build() has just re-authored the bar, so whatever folding state a PREVIOUS
     // bar was in no longer describes this DOM. Without this reset a rebuild on a
     // phone (switching language, say) would early-return on `true === true` and
     // leave the freshly authored DESKTOP bar in place — overflowing, with Save
     // off-screen again.
     this.phoneChromeOn = null
-    this.applyPhoneChrome(this.phoneQuery.matches)
-    this.phoneQuery.addEventListener('change', (e) => this.applyPhoneChrome(e.matches))
-    // ...and on plain resize as well. matchMedia's change event is the correct
-    // signal but not a universally reliable one — it does not fire at all under
-    // CDP-driven viewport changes, and a phone ROTATING is exactly this path.
-    // applyPhoneChrome early-returns when the state is unchanged, so calling it
-    // on every resize costs a comparison.
-    window.addEventListener('resize', () => this.applyPhoneChrome(window.innerWidth <= 700))
+    this.topbar = bar
+    this.fitTopbar()
+    // A ResizeObserver on the bar itself is the primary width signal. It fires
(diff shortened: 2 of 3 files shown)


bouzidanas:
I should add that while this fix will add some complexity to maintaining/reviewing code (because layout change logic will be in js), it is more scalable. Don't need to redo breakpoints if anything is added/changed/removed from the top bar.

nyblnet:
Rebased onto `main` and CI is green — thank you, and sorry it sat without a run for so long: the branch predated a workflow change, so nothing had ever triggered.

**The review came out well.** I checked the thing I was most worried about — #212 fixed a bug where unfolding the topbar restored buttons by a rule rather than by memory, so Comment migrated out of the insert group and Save ended up after Help. Your change leaves `applyPhoneChrome`'s `authored`/`homeOf` replay untouched and only changes *who calls it*. Verified with a real fold→unfold cycle in a browser, not just by reading the diff: the restored bar comes back byte-identical to the authored order.

The "hiding text, never scrolling" rule is honoured, both fold call sites still go through `applyPhoneChrome`, and the observer pair doesn't feed back on itself — `class` is excluded from the `attributeFilter`, and repeated runs go static rather than oscillating.

**I'm holding it out of the imminent release for one thing**, which is small:

`fitTopbar` early-returns while a dropdown is open, and closing a dropdown doesn't schedule a refit (`class` isn't observed). So: open ⋯ on a folded bar → widen the window → close ⋯, and the bar stays folded until the next resize or content change. Self-correcting, but visible. A refit scheduled when a dropdown closes would clear it.

Two more for your judgement rather than blockers:

- The ResizeObserver watches the default content-box while the tier classes change the bar's own padding (14 → 8 → 6px), so the callback resizes what it observes — the classic shape behind *"ResizeObserver loop completed with undelivered notifications"*. It converges and I didn't see the warning, but `observe(bar, {box: 'border-box'})` would remove the question.
- Pre-existing, not yours: the `window.resize` listener in `build()` is never removed, so a language switch stacks another `fitTopbar` per resize event.

The next release is a security one, so I'm keeping it to the fix plus small, contained changes — that's the only reason this is waiting. It goes into the following release, which is the feature one.

nyblnet:
I've pushed two things to this branch — sorry for touching it, but one of them only exists *because* of the class-based approach and I'd rather it not bite you.

**Rebased onto current `main`** (it now includes #264 and #258; #258 changes topbar content, which your MutationObserver reacts to, so it's worth having underneath).

**Aligned the tight tier's padding with a fix going into `main` separately** (#284). We noticed the bar changes height as a window is dragged — 47px, then 43px below 760, then 57px below 700. The 43px step came from dropping vertical padding 8 → 6, which reclaimed 4px of canvas and cost a visibly moving toolbar. The horizontal 14 → 6 is untouched; that one earns its keep.

So the tight tier now narrows horizontally only, and the 6px vertical moves to the phone rule where the 44px touch targets already set the height.

**The bit specific to your change:** in the phone media query a bare `.ed-topbar` (0,1,0) loses to `.ed-topbar.ed-bar-tight` (0,2,0), because a media query adds no specificity. So the phone bar silently kept desktop padding — measured 61px instead of 57px. I've named `.ed-bar-tight` explicitly there. That trap doesn't exist on `main`, where both sides are media queries; it only appears once the tiers become classes, so it's worth knowing about for any future tier rule that needs overriding on phones.

**Also, from testing in a real browser:** the ResizeObserver does fire properly (it never did in the headless environment we first checked in), all three tiers apply cleanly, no loop warning appeared, and the restore order after unfolding is byte-identical — Comment stays in the insert group, Save stays before Help.

One thing I noticed and haven't chased: resizing 640 → 740 left `ed-bar-fold` applied, where loading fresh at 740 doesn't fold. Might be legitimate hysteresis near the threshold, might be the same staleness as the dropdown case. Worth a look when you pick up the refit-on-dropdown-close change.

CI will need approving again after this push.

PR #259, opened 2026-08-07
Feature: Code snippets in Slides with morph (demo attached)

tikurahul:
**What & why**

  Code snippets with morphing in slides. 

  * Define a `CodeElement`, panels and render code in slides.
  * `CodeElement`s `morphId` defines the group of slides that are diffed. So, we **never** diff more than we need. 
  * The diff algorithm used is linear time, and is really slick. It's inspired by [this](http://portal.acm.org/citation.cfm?id=359460.359467&dl=GUIDE&dl=ACM&idx=359460&part=periodical&WantType=periodical&title=Communications%20of%20the%20ACM) paper.
  * Make `Edit` mode work gracefully.
  * Fully automated morphing using the existing animation engine (uses the ideas from MathML rendering).
  * The model **never** stores the HTML. It only stores the raw text, and therefore versioning is rather simple. 
  * Reduced runtime costs for Shiki by only using the Parser, and stripping out everything else.
  
**How I verified it**
  * `npm run build:single`
  * Attached are the demos.

**Before**
```
[plugin vite:singlefile] Inlining: index-BNdNa_z8.js
[plugin vite:singlefile] Inlining: style-C0jTj-fw.css
dist-single/index.html  1,477.10 kB │ gzip: 513.40 kB
✓ built in 1.50s
compressed shell: 1379KB → 674KB (js 1276KB→636KB, css 97KB→27KB)
```

**After**
```
[plugin vite:singlefile] Inlining: index-_qoGBxdk.js
[plugin vite:singlefile] Inlining: style-B6335lfm.css
dist-single/index.html  1,651.73 kB │ gzip: 569.62 kB
✓ built in 1.80s
compressed shell: 1549KB → 747KB (js 1446KB→709KB, css 97KB→28KB)
```

**Checklist**
- [x] Ran `node scripts/test-sync.ts` if I touched `slides/src/sync/`
- [ ] Document format changes are additive and backward-compatible
- [ ] Did not bump the version or cut a release (maintainers sign releases)

### DIFF
--- a/slides/src/code.ts
+++ b/slides/src/code.ts
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 The Bento authors
+
+import { ThemedToken } from 'shiki'
+import { Element } from 'hast'
+import { BentoDoc, CodeElement } from './model';
+import { sanitizeHtml } from './render';
+import { getOrCreateHighlighter, HeckelDiff, Insert, State, Token, Delete, Match } from './diff';
+
+let diff: HeckelDiff | undefined = undefined
+
+export type RenderCodeResult = {
+  /** The HTML representation when the `CodeElement` has a `grammar` and a `theme` defined. */
+  html?: string,
+}
+
+/**
+ * This returns the HTML representation of the `CodeElement` when the corresponding `grammar` and `theme` is defined.
+ * Otherwise, it returns undefined.
+ */
+export function renderCode(el: CodeElement, doc: BentoDoc) {
+  const mid = el.morphId
+  let sanitized: string | undefined = undefined
+  if (el.grammarAssetId && el.themeAssetId && el.grammarName && el.themeName) {
+    const slideIdx = slideIndex(el, doc)
+    const diff = getOrCreateDiff(doc)
+    const map = diff.computeDiffs(mid)
+    const states: State[] = map.get(slideIdx) ?? []
+    const highlighter = getOrCreateHighlighter(doc, el.grammarAssetId, el.themeAssetId)
+    const html = highlighter.codeToHtml(el.content, {
+      // Bad typings here
+      lang: el.grammarName as any,
+      theme: el.themeName as any,
+      includeExplanation: 'scopeName',
+      mergeSameStyleTokens: false,
+      transformers: [
+        {
+          // @ts-expect-error : We don't really need col, and lineElement
+          span(hast: Element, line: number, col: number, lineElement: Element, token: ThemedToken) {
+            // 0-indexed line number
+            const lineNumber = line - 1
+            // There should always be a matching state, but if we do intermediate renders then we
+            // should still gracefully recover.
+            const state = findMatchingState(states, lineNumber, token)
+            if (state) {
+              const id = morphId(state)
+              hast.properties['data-flip-id'] = id
+            } else {
+              console.warn('Inconsistent States:', states, 'Line Number:', lineNumber, 'Token:', token)
+            }
+          },
+        }
+      ]
+    })
+    // sanitizeHtml(...) strips away all the skiki background styles.
+    // That part, is great given the background no longer interferes with rendering.
+    sanitized = sanitizeHtml(html)
+  }
+  return {
+    html: sanitized
+  }
+}
+
+function getOrCreateDiff(doc: BentoDoc): HeckelDiff {
+  if (!diff) {
+    diff = new HeckelDiff(doc)
+  }
+  return diff
+}
+
+/** Computes the slide index for the `CodeElement` being rendered. */
+function slideIndex(el: CodeElement, doc: BentoDoc): number {
+  const slides = doc.slides
+  for (let i = 0; i < slides.length; i += 1) {
+    const slide = slides[i]
+    const filtered = slide.elements.filter(element => element === el)
+    if (filtered && filtered.length > 0) return i
+  }
+  // Should never really happen
+  return -1
+}
+
+function findMatchingState(states: State[], lineNumber: number, themedToken: ThemedToken): State | undefined {
+  const token = Token.fromThemed(lineNumber, themedToken)
+  const state = states.find((state) => {
+    const kind = state.kind
+    if (kind === 'empty') {
+      return false
+    } else if (kind === 'insert') {
+      const insert = state as Insert
+      return insert.token.key() === token.key() && insert.token.lineNumber === token.lineNumber && insert.token.offset === token.offset
+    } else if (kind === 'delete') {
+      const del = state as Delete
+      return del.token.key() === token.key() && del.token.lineNumber === token.lineNumber && del.token.offset === token.offset
+    } else if (kind === 'match') {
+      const match = state as Match
+      return match.current.key() === token.key() && match.current.lineNumber === token.lineNumber && match.current.offset === token.offset
+    }
+    return false
+  })
+  return state
+}
+
+function morphId(state: State): string {
+  let id = ''
+  if (state.kind === 'insert') {
+    const insert = state as Insert
+    id = insert.token.morphId()
+  } else if (state.kind === 'delete') {
+    const del = state as Delete
+    id = del.token.morphId()
+  } else if (state.kind === 'match') {
+    const match = state as Match
+    id = match.current.morphId()
+  }
+  return id
+}
--- a/slides/src/code.ts
+++ b/slides/src/code.ts
@@ -1,117 +1,104 @@
 // SPDX-License-Identifier: MIT
 // Copyright (c) 2026 The Bento authors
+//
+// Render bridge for code elements: kernel tokens -> HeckelDiff identities ->
+// morphable spans. The diff engine (diff.ts) decides WHICH token on slide N is
+// the same token as on slide N-1; this file only turns that answer into DOM.
+//
+// Tokens are emitted as `data-sym` spans inside the element's own
+// `data-flip-id` host — the same sub-element channel the formula morph uses —
+// NOT as per-token flip ids. The morph engine is model-driven: a flip id with
+// no model entry is skipped by every matched-pair consumer, so per-token flip
+// ids animate insertions (the entering path needs no model) but let matched
+// tokens TELEPORT. data-sym offsets are measured per host on slide entry and
+// tweened on exit (present.ts cacheSlideSymbols/morphMathSymbols), which is
+// what makes a moved line travel.
 
-import { ThemedToken } from 'shiki'
-import { Element } from 'hast'
-import { BentoDoc, CodeElement } from './model';
-import { sanitizeHtml } from './render';
-import { getOrCreateHighlighter, HeckelDiff, Insert, State, Token, Delete, Match } from './diff';
+import { BentoDoc, CodeElement } from './model'
+import { HeckelDiff, type Match, type Insert, type Delete } from './diff'
 
-let diff: HeckelDiff | undefined = undefined
-
-export type RenderCodeResult = {
-  /** The HTML representation when the `CodeElement` has a `grammar` and a `theme` defined. */
-  html?: string,
+/**
+ * Seven colours, not four hundred scopes — the zero-cost tier. When the
+ * signed-extension tier lands, grammarAssetId/themeAssetId select real
+ * TextMate rendering and this map becomes the fallback.
+ */
+const CODE_COLORS: Record<string, string> = {
+  c: '#6b7f8f', // comment
+  s: '#c98a3e', // string
+  n: '#b0688f', // number
+  k: '#5b8def', // keyword
+  f: '#3fa9a0', // call
+  p: '#7c8794', // punctuation
+  a: '#3f9142', // diff: added
+  d: '#c25a43', // diff: removed
+  x: '',        // plain: inherit the element colour
 }
 
 /**
- * This returns the HTML representation of the `CodeElement` when the corresponding `grammar` and `theme` is defined.
- * Otherwise, it returns undefined.
+ * Memoized per (morph group, content). The previous module-level singleton was
+ * cached forever — built from the first document it saw and never invalidated,
+ * so live edits kept diffing yesterday's content. Identity-keyed caching does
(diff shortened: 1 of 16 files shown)


tikurahul:
To try the demo end to end:

* Cherry pick the patch. 
* Add a `CodeElement` to the slide.  Add a `morphId` and select `Morph` as your transition.
   - Pick the `TextMate` grammar to use from [here](https://github.com/shikijs/textmate-grammars-themes/tree/main/packages/tm-grammars/grammars) repo. (Download and attach).
   - Pick the `TextMate` theme you want from [here](https://github.com/shikijs/textmate-grammars-themes/tree/main/packages/tm-themes/themes).
* Clone the slide and make changes to your code.
* The demos look a lot nicer than V1 (the first demo I attached). Will add another one.

tikurahul:
Demo with prettier animations: 
[Bento-Code-Demo-V2.webm](https://github.com/user-attachments/assets/bae618ba-affc-4833-928d-359aefbc5b6d)

tikurahul:
Made the demos look evern cooler 😄 
Now, the PR is ready, and I don't any changes to add (apart from addressing your feedback i.e.).

slides/src/code.ts
```diff
@@ -0,0 +1,117 @@
...
+            // 0-indexed line number
+            const lineNumber = line - 1
+            // There should always be a matching state, but if we do intermediate renders then we
+            // should still gracefully recover.
+            const state = findMatchingState(states, lineNumber, token)
+            if (state) {
+              const id = morphId(state)
+              hast.properties['data-flip-id'] = id
+            } else {
+              console.warn('Inconsistent States:', states, 'Line Number:', lineNumber, 'Token:', token)
```
tikurahul: Note to reviewer: 

This is something that should never really happen. This is useful context when filing a bug. If you have ideas on how else to log this kind of diagnostic information, please let me know.

nyblnet:
Thanks for submitting this. It's summer holidays so I'm a bit behind on dealing with the PRs but will take a look at this and how to slim down the runtime next.

tikurahul:
Thanks for getting back. Please take your time and enjoy the holidays 😊

nyblnet:
Spent a while with this one, and I want to flag something in your diff that looks cosmetic and is actually load-bearing:

```css
.bento-el-code span {
  display: inline-flex;
}
```

I reimplemented the render path locally to measure the tokenizer cost, didn't carry that rule across, and lost about a day to it. A CSS transform does **nothing** to a non-replaced inline element — the browser accepts the declaration, `style.transform` and `getComputedStyle` read it back verbatim, and the box moves zero pixels. Silently. So the morph reported perfect health while painting absolutely nothing:

```
1→2  ANIMATED: 20 tokens, 199 frames, ended p=0.939
animation clock   120 fps
tab visibility    visible
reduced motion    off
```

Every check I wrote read back `style.transform` — a value I'd just written myself — so it passed identically whether the tokens travelled 400px or none. Only `getBoundingClientRect` could ever have caught it. Your `inline-flex` is what makes the whole feature work, and it's one `git blame`-less cleanup away from being deleted as redundant styling.

Two asks, both small:

1. **A comment on that rule** saying transforms are ignored on inline elements and the morph dies silently without it. It reads like formatting; it's the load-bearing line.
2. **Whitespace tokens** — worth checking how yours behave. I ended up scoping the display change to non-whitespace tokens only, because a token carrying a newline inside `white-space: pre` makes an inline-level box multi-line and disturbs the layout. May not apply to how you tokenize, but it bit me.

I've also put up #361 on our side: `kernel/src/anim.ts` will now warn once when a transform channel targets an inline element, so the next person gets a console message instead of a day. MathML tokens compute to `block math` and are unaffected, and replaced elements and SVG children are exempt — measured, not assumed.

Separately, on the tokenizer question from earlier: I built a comparison and the numbers hold up well for your approach. More on that in a follow-up, along with a couple of things I found about morph tempo — the default 0.65s with a `power2.inOut` ease is tuned for elements travelling hundreds of pixels, and a code token moving one line-height reads as a blink at that speed.

nyblnet:
Following up on my note above — I measured this properly and I got the mechanism wrong the first time. Correcting it, because if you go looking for a fix based on what I said you'd look in the wrong place.

**It is not about `inline-flex`.** I assumed flex layout was the risk. It isn't. Applying an atomic inline-level display to *every* token span collapsed a nine-line block to one line and shifted x-positions (139→132, 199→192) — and it did that **identically under `inline-block` and `inline-flex`**.

Isolating which spans cause it:

| spans given a box | lines rendered |
|---|---|
| ink-bearing tokens only | 9 ✓ |
| \+ space-only tokens (`" "`) | 9 ✓ |
| \+ tokens containing `\n` | **1** ✗ |

The culprit is a token that *carries a newline*. Inside any atomic inline-level box, that `\n` stops breaking the line in the parent's `white-space: pre` flow. The display value is irrelevant — what matters is never blockifying a token containing a line break.

So the question for `.bento-el-code span` is only this: **does your tokenizer ever emit a span containing `\n`?** If shiki gives you line structure separately and newlines never land inside a token span, your rule is safe as written and simpler than what I ended up with — ignore all of this. If newlines do ride inside spans, it's a latent layout bug your demo may not exercise, and the fix is scoping rather than changing the display value.

For what it's worth I landed on `.bento-code [data-tok] { display: inline-block }`, with the renderer setting `data-tok` only on tokens with ink. One rule rather than a per-span inline style, and it keeps ~19 style attributes per code block out of our static first-page preview, which is byte-budgeted. Same idea as yours, just scoped.

The "add a comment to that rule" ask from my previous message stands, and honestly matters more now: whichever form it takes, it reads like cosmetic styling and it is the line that makes the feature work.

nyblnet:
I've rebased this onto current main so it's mergeable again — only `package-lock.json` conflicted, and I regenerated it from your `package.json` rather than hand-resolving. Your commit and authorship are untouched (`82cb18b`, still yours). `tsc` clean, builds, boots with no console errors. Plan is to merge it as your commit rather than cherry-pick pieces of it.

First, a correction I owe you. I said above it'd be worth checking whether your diff spreads matches in both directions — I'd measured that as worth ~8 points of match rate in my own implementation. **You already do it.** Phase 3 is the backward pass, right there next to Phase 2. I'd written my own Heckel pass before properly reading `diff.ts`, and then reported a gap that doesn't exist. Having now read it: your four-phase structure with the explicit `Match`/`Insert`/`Delete` state is cleaner than what I wrote, and it's the version that should survive.

**What I'd like to build on top, and why it isn't a replacement of your work.**

The thing I keep coming back to is that your format is already tiered, and I don't think that was an accident:

```ts
grammarAssetId?: string
themeAssetId?: string
```

Optional, grammars and themes carried as per-deck **assets** rather than baked into the shell, and `renderCode` returning `undefined` when they're absent. That's a seam, and it's the right one.

So what I'd add is a second tier *behind* that seam: a small built-in tokenizer (~6.7KB, 77 languages, no assets, no dependency) that handles the `grammarAssetId`-absent branch. Rough syntax for the common case at zero per-deck cost; shiki when a deck wants real grammars and is willing to carry them. Your `diff.ts` sits underneath both — it's language-agnostic, it doesn't care where tokens come from.

Measured cost of the shiki engine in the shell, now that I can build it: **652KB → 749KB compressed, so +97KB.** Worth knowing because Bento Spaces has a ~140KiB budget where that can't fit — but that's a *future* question about where the engine lives, and your asset-based design means answering it later needs **no format change**. Nothing is locked in by merging now.

Two small things I'd still like on the diff before it goes in:

1. A comment on `.bento-el-code span { display: inline-flex }` marking it load-bearing (context in my earlier messages — it reads like formatting and the feature dies silently without it).
2. Nothing else blocking.

On the newline question from my last comment: I checked, and your transformer is per-line (`span(hast, line, col, lineElement, token)`), so shiki hands you line structure and newlines never land inside a token span. Your rule is safe as written. That whole concern was specific to my tokenizer, which emits newline-bearing tokens — my problem, not yours, and it'll ride along with that change rather than landing on you.

Thanks for this — the diff engine is the hard part and you got it right.

nyblnet:
An update with a decision in it, and the reasoning laid out in full since it changes the shape of this PR.

**The size line.** We measured the cost precisely by building this branch with shiki stubbed out: your scaffolding — the diff engine, the `CodeElement` format, panel/canvas/editor wiring — is **+27KB** compressed. The shiki engine itself is **+70KB** on top (652KB → 749KB shell). Bento's core promise is that every saved deck carries the whole runtime, so that 70KB lands in every file ever saved, code or no code. That's over the line for us — not a judgement on shiki, which is the right tool in almost any other architecture. It's specifically the self-contained-file constraint biting.

**But while verifying that, we found something that matters more than size.** I traced how the current morph engine treats your token spans, end to end. `runMorph` is deliberately model-driven — both frames come from the document, no DOM measuring — and every consumer of a matched `data-flip-id` pair does a model lookup and skips when it misses. Your minted token ids (`m-code-token-N`) have no model entry, so:

- **inserted tokens animate** — the entering path needs no model, and your `inline-flex` is what makes those spans transformable, which is why your fill-in demo looks as good as it does;
- **matched tokens that change position teleport** — the travel case structurally can't reach them.

So a reorder — the "two lines swap places" case — snaps rather than travels, and that's an engine constraint no highlighter choice can fix. The sanctioned path for sub-element motion is the `data-sym` mechanism the formula morph uses (measured travelling 399px in a test deck this week, after we fixed a separate bug where inline tokens silently ignored transforms — your `inline-flex` had already dodged that one).

**What we'd like to do, with your blessing:**

- **Merge your commit as-is, authorship intact** — it's already rebased and green.
- Follow-up commits (clearly ours, on top of yours) that: feed your `HeckelDiff` from a small built-in tokenizer (~7KB, 77 languages, type+text keys — a graceful degrade from your scope-array keys) instead of `Token.fromThemed`; emit tokens through the `data-sym` path so moves actually travel; and take shiki out of the bundle.
- **Your format fields stay exactly as designed.** `grammarAssetId?`/`themeAssetId?` remain in the model — they're the seam a future signed-extension mechanism lights up, with `code.ts` in this PR's history as its reference implementation. Grammars-as-deck-assets was the right call and it's what makes deferring the engine cheap.
- One small fix riding along: the module-level `HeckelDiff` singleton never invalidates when the doc changes — fine in present mode, stale in the live editor.

Net: the feature ships at ~34KB instead of 97KB, insertions keep the behaviour your demo shows, moves gain real travel, and your diff core — the hard part — is the engine under all of it. The built-in tokenizer lives in the shared kernel, so Spaces (which has a ~140KiB total budget) gets code morphs too, which shiki could never have fit.

If any of this reads wrong to you, say so before we push — it's your branch and your feature. And genuinely: the four-phase diff with morph-id propagation held up to close reading better than my own reimplementation of it did.

nyblnet:
Correcting a number from my long comment above, in your favour. The "+27KB for your scaffolding" was a bad measurement — the two builds sat on different commits of `main`, and unrelated PRs landing in between got attributed to your code. Re-measured with every variant on the same base:

| | delta |
|---|---|
| your scaffolding — diff engine, format, panel/canvas/editor UI | **+2KB** |
| of which the Heckel core | <1KB |
| shiki engine | +70KB (unchanged) |

So the feature you built is two kilobytes of shell; effectively everything was the highlighter engine. The plan from the earlier comment is unchanged — it was always about the 70KB — but the record should show your code was never a meaningful part of the size question.

tikurahul:
Thanks for an **extremely** thorough review. 

* The `inline` elements not being animated indeed is **load bearing**. I used `inline-flex` purely for convenience, but it could just have easily been an `inline-block` as well. (I lost about a day to it too 😄 ). 

* The `data-sym` path sounds incredible. That is the one part I wanted to follow up on, because the positional changes for matched elements, did not behave as expected and I assumed I was holding something wrong.

* I always thought we might eventually want to drive a lot of this by writing our own tokenizer. So, I wrote all the diffing and animation assuming that in the future if we want to go down that way - that should be a trivial change. So the fact that you already do something for basic language support is pretty amazing.

--

All your other suggestions SGTM. If there is anything you would like me to help with or follow up on, please let me know.

nyblnet:
Merged — thanks for this one, it's a genuinely great feature and it made the cut list for the next release.

Your commit went in as-is, with the follow-ups stacked on top: your diff engine now takes tokens from the small built-in tokenizer, shiki is out of the shell, and the grammar/theme asset fields are untouched, waiting for a proper extension mechanism to light them up. You called it in your reply — you'd built the diffing assuming a homegrown tokenizer might slot in one day, and it slotted in exactly as you predicted. The whole feature lands at about 10KB.

One bonus that fell out along the way: a swap of two lines sharing a prefix (two `const` lines trading places) used to be undiffable — the second line's tokens lost their identity and blinked. The spread phases now work outward from the anchors ring by ring, so each token pairs with the nearest anchor and travels with its own line. There are regression tests in scripts/test-codediff.ts covering that and a few smaller pairing edge cases we hit during review.

Since the tokenizer lives in the shared kernel, code blocks in the upcoming notes app can get the same treatment, which shiki's size would never have allowed. Your diff engine will be doing the work there too.

Happy to have you around — if the extension mechanism ever materialises, your original shiki binding is the reference implementation, and you'd be the obvious person to bring it back.

tikurahul:
> One bonus that fell out along the way: a swap of two lines sharing a prefix (two const lines trading places) used to be undiffable — the second line's tokens lost their identity and blinked. The spread phases now work outward from the anchors ring by ring, so each token pairs with the nearest anchor and travels with its own line. There are regression tests in scripts/test-codediff.ts covering that and a few smaller pairing edge cases we hit during review.

Great catch, and really good fix!

> Since the tokenizer lives in the shared kernel, code blocks in the upcoming notes app can get the same treatment, which shiki's size would never have allowed. Your diff engine will be doing the work there too.

This is **really** amazing. I hoped that I could make a small difference to this incredible project, and am very thrilled that I can. Your original `Show HN` post really made me wonder if I could come up with an elegant implementation to solve for this key usecase.

> Happy to have you around — if the extension mechanism ever materialises, your original shiki binding is the reference implementation, and you'd be the obvious person to bring it back.

Yes! I will be happy to bring it back. 

Just wanted to add that I am also excited to play around with all the bits of polish you added to the original PR.

PR #271, opened 2026-08-10
slides: a tap opens the text box a tap selected — text is uneditable on touch

iamgeo92:
## The problem

**On a phone or tablet, a bento deck is effectively read-only.** You can select
elements, drag them, resize them — but you cannot change a single word.

Opening a text box is bound to one event:

```ts
// slides/src/editor/canvas.ts
this.stage.addEventListener('dblclick', (ev) => { … startTextEdit(textEl) … })
```

`dblclick` never reaches the canvas from a touch screen. Selecto and Moveable
call `preventDefault()` on the touch stream they handle, and that suppresses the
*entire* synthesized mouse sequence — `click` and `dblclick` alike. Instrumenting
the canvas under an iPhone profile, a double-tap on a text element delivers:

```
pointerdown, touchstart, touchend      ← and nothing else
```

There is no second route into `startTextEdit`, so there is no way in.

## The fix

A tap on the element that a **previous tap already selected** opens it — the
gesture every touch editor uses for "open this". A table opens the cell under
the thumb.

Two details carry the change:

- **The "was it selected?" question is asked on the press, not the release.**
  Selecto selects on the press, so by the time the finger lifts, the tap that
  merely *selected* has already made the element the selection and would open
  the editor on the **first** tap. Asking on `touchstart` is what keeps
  tap-to-select intact.
- **The opening tap is cancelled, so the compatibility mouse burst is never
  synthesized.** A touch the browser does not see cancelled is replayed as
  `mousedown → mouseup → click` ~300ms later at the *original* screen point —
  which `focus()` has by then scrolled away from, to bring the caret into view.
  Selecto read that ghost press as "pressed outside the text being edited" and
  committed the edit a blink after it opened. This only reproduced once the
  canvas was zoomed enough to scroll, which is exactly the state a phone is in.
  Only the tap that actually opens an editor is cancelled; a tap that merely
  selects is left alone.

Hit-testing goes through `document.elementsFromPoint` rather than the model
boxes, because that is the same question Selecto's own click selection answers —
"what is painted here?". A model-box test is *not* equivalent: decorative
full-bleed elements overlap smaller ones in model space while sitting behind
them on screen, and the two disagreed on the starter deck's own headline.

## Before / after

Identical gesture sequence in both panels: tap, tap again, then double-tap.

| before | after |
|---|---|
| tap, tap, double-tap — nothing opens | second tap opens the box; typing works |

tap to edit, before and after

<details>
<summary>Stills</summary>

| before — no caret, no keyboard | after — edited and committed |
|---|---|
|  |  |

</details>

## Nothing changes for a mouse

- Single click still selects; it can never open, because the selecting tap is
  never the opening one.
- Double-click still opens, unchanged.
- The recogniser is bound to **touch events**, which cannot fire for a mouse at
  all — so there is no `pointerType` filter to get wrong, and nothing on a
  hybrid device (a touchscreen laptop, an iPad with a trackpad) can reach it
  through the mouse.
- Nothing listens for mouse events on the document, and nothing is suppressed
  after the fact.

## How this was verified

`slides/` on the dev server, driven through Chrome DevTools' iPhone emulation
(390×664, DPR 3, `hasTouch`), with real touch event dispatch — plus a mouse
control at 1280×800 for the regression cases:

| case | result |
|---|---|
| tap #1 selects, does not open | pass |
| tap #2 on the selected box opens it | pass |
| typed text reaches the document model | pass |
| a drag (finger moves >10px) does not open | pass |
| a long press (>700ms) does not open | pass |
| tapping a *different* element re-selects, never opens | pass |
| table: tap opens exactly the cell under the thumb | pass |
| desktop double-click still opens | pass |
| a slow second mouse *click* does not open | pass |
| zoomed canvas (the scrolling case) opens and stays open | pass |

Gates run on this branch:

```
```

No new user-facing strings, so no i18n catalog changes. No format change, no
kernel change, nothing under `sync/`.

## Update after review

@7jameslondon built and tested the first version on a **real iPhone** and on
desktop, and suggested recognising the gesture on touch events and cancelling
the tap rather than tracking pointer events and filtering the mouse events they
produce. That is a better shape and it is now what the branch does — second
commit on this PR.

The first version needed *two* mechanisms to protect desktop behaviour: a
`pointerType === 'touch'` guard, and a window-level capture listener that
swallowed `mousedown`/`mouseup`/`click` for 450ms within 25px of the touch.
Both were mine to get wrong. Cancelling the `touchend` stops the browser
generating that burst in the first place, so `swallowGhostMouse` and its two
constants are gone — a net **−21 lines**, and the guard is structural rather
than heuristic.

Two assertions were added with it: the opening tap produces **no** synthesized
mouse events, and the *selecting* tap is **not** cancelled.

**Verified on a real iPhone.** My own testing is Chromium's iPhone emulation, so
I flagged one thing emulation cannot prove: that the on-screen keyboard still
opens when the tap that focuses the text is also the tap that gets cancelled.
@7jameslondon has since re-tested **this** commit on a real iPhone and on
desktop and confirmed it — *"everything works as expected including the keyboard
hiding and showing only when it should"* — so that caveat is closed.

---

<details>

```markdown
- **Fix: text could not be edited at all on a phone or tablet.** Opening a text
  box was bound to `dblclick`, and that event never reaches the canvas from a
  touch screen — Selecto and Moveable call `preventDefault` on the touch stream
  they handle, which suppresses the whole synthesized mouse sequence, `click`
  and `dblclick` alike. There was no second route, so on an iPhone the deck was
  effectively read-only: elements selected, moved and resized, but no words
  could be changed. A tap on the element a previous tap already selected now
  opens it — the gesture every touch editor uses for "open this" — and a table
  opens the cell under the thumb. Nothing changes for a mouse: the tap that
  selects can never be the one that opens, so single-click selection is intact,
  and double-click still works as before.
```

</details>

### DIFF
--- a/slides/src/editor/canvas.ts
+++ b/slides/src/editor/canvas.ts
@@ -26,11 +26,6 @@ import type { Peer } from '../sync/session'
 const TAP_SLOP = 10
 /** …and how long it may rest. Past this it is a press, not a tap. */
 const TAP_HOLD_MS = 700
-/** The compatibility mouse burst follows a touch by ~300ms; 450 covers a slow
- *  frame without reaching far enough to catch a deliberate second interaction. */
-const GHOST_WINDOW_MS = 450
-/** …and it lands within a pixel or two of the touch, never across the screen. */
-const GHOST_SLOP = 25
 
 export class SlideCanvas {
   private stage: HTMLElement
@@ -274,13 +269,24 @@ export class SlideCanvas {
     // a previous tap already selected, so that is what enters editing. Checking
     // the selection first is what keeps it safe: the tap that SELECTS an element
     // can never be the one that opens it, so nothing is swallowed.
+    //
+    // TOUCH events, not pointer events, and deliberately so. A pointer handler
+    // runs for a mouse as well and has to filter itself back out by
+    // pointerType; the compatibility mouse burst it leaves behind then has to
+    // be filtered too — two separate chances to change desktop behaviour by
+    // accident. A touch handler cannot fire for a mouse at all, and cancelling
+    // the touchend stops the browser SYNTHESIZING that burst in the first
+    // place, so there is nothing left to filter. (Suggested by @7jameslondon,
+    // who tested the pointer version on a real iPhone.)
     let tap: { x: number; y: number; t: number; id: string | null; wasSelected: boolean } | null = null
-    this.scroller.addEventListener('pointerdown', (ev) => {
-      if (ev.pointerType !== 'touch' || !ev.isPrimary) { tap = null; return }
-      const id = this.topElementAt(ev.clientX, ev.clientY)
+    this.scroller.addEventListener('touchstart', (ev) => {
+      // a second finger means a pinch, never a tap
+      if (ev.touches.length !== 1) { tap = null; return }
+      const t = ev.touches[0]
+      const id = this.topElementAt(t.clientX, t.clientY)
       const sel = this.store.selection
       tap = {
-        x: ev.clientX, y: ev.clientY, t: ev.timeStamp, id,
+        x: t.clientX, y: t.clientY, t: ev.timeStamp, id,
         // Asked HERE and not on release, because Selecto selects on the press:
         // by the time the finger lifts, the tap that merely selected has
         // already made this element the selection and would open the editor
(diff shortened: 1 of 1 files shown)


7jameslondon:
This is a great fix! And I really appreciate the before and after demo. I have built and tested this PR myself on a desktop computer with mouse usage and on a real iPhone with touch usage. In both cases everything looks pretty good.

However, I would suggest that instead of tracking pointer events and then filtering the mouse events they create, track the touch events directly (`.addEventListener('touchstart', ...`), filter out movement and long presses, and prevent the valid touch gesture from creating mouse events in the first place. This would even further reduce the risk of changing mouse behavior on desktop while supporting iOS browsers.

iamgeo92:
Thank you — and thank you especially for building it and putting it on a real iPhone. That was the one gap I flagged in the description and couldn't close myself.

You're right, and I've taken it: [`7892c77`](https://github.com/nyblnet/bento/commit/7892c77) moves the recogniser onto `touchstart`/`touchmove`/`touchend` and cancels the tap that opens an editor, so the compatibility burst is never generated instead of being raced afterwards.

It's a clear improvement, and concretely so — the first version needed **two** mechanisms to protect desktop behaviour:

1. a `pointerType === 'touch'` guard on a handler that runs for a mouse as well, and
2. a window-level capture listener swallowing `mousedown`/`mouseup`/`click` for 450 ms within 25 px of the touch.

Both were mine to get wrong. A touch handler can't fire for a mouse at all, so (1) becomes structural rather than a filter, and cancelling the `touchend` removes the need for (2) entirely. `swallowGhostMouse` and its two constants are gone: **net −21 lines**, and nothing in the editor listens for mouse events on the document any more.

I checked the premise before rewriting rather than assuming it — on the canvas the order is `pointerdown → touchstart → pointerup → touchend`, the `touchend` is `cancelable`, and cancelling it does suppress the burst. Two assertions came with the change:

- the opening tap produces **no** synthesized mouse events (`[]`, previously the ghost `mousedown` that was committing the edit); and
- the tap that merely **selects** is *not* cancelled — only the one that actually opens an editor is, so ordinary tap-to-select keeps its default behaviour.

The rest of the suite still passes, including the case that motivated the ghost filter in the first place (a zoomed canvas, where `focus()` scrolls the slide out from under the original touch point), plus table cells, drag, long-press, and the two desktop regressions.

One thing worth naming, since your device test was of the *previous* commit: the behaviour I can't prove in emulation is that the on-screen keyboard still opens when the focusing tap is also the cancelled one. `focus()` is still called synchronously inside the gesture, which is what iOS requires, and cancelling `touchend` shouldn't affect that — but "shouldn't" is doing real work in that sentence. If you have a moment to re-run your iPhone test on the new commit, that's the assertion I'd most like a second pair of eyes on.

I also confirmed it composes with #278 (pinch-to-zoom), which adds its own touch listeners to the same scroller: this handler bails on `touches.length !== 1`, both suites pass on the merged tree, and the only collision is the CHANGELOG one.

7jameslondon:
I tested the updated PR on a real iPhone and desktop. Everything works as expected including the keyboard hiding and showing only when it should. At the moment it is still a pain to try to do anything on iOS in a browser but I think your other PRs will help that a lot. When I have time I will try to give those a test as well.

For what it is worth @nyblnet I think this PR is ready to merge and is a good addition to the project.

iamgeo92:
Thanks, I made some more PRs to make the experience on mobile better. Would love to hear your thoughts on those too.

iamgeo92:
@nyblnet — when you have a moment, could I ask you to take a look at this one?

It's the headline fix of a batch of mobile work: text was **completely uneditable** on a phone or tablet, because opening a box was bound to `dblclick` and that event never reaches the canvas from a touch screen. A deck on an iPhone was effectively read-only.

@7jameslondon has already built and tested it on a real iPhone (thank you!) and suggested moving the recogniser onto touch events and cancelling the tap at source rather than filtering the mouse events afterwards — that's the second commit, and it deleted more than it added.

<details>
<summary><b>The other seven, and a suggested order</b> — all mobile/editor, all independent unless noted</summary>

`docs/PARALLEL-WORK.md` §2 says independent things first, then the overlapping ones one at a time, so:

| | PR | |
|---|---|---|
| 1 | #271 | text is uneditable on touch ← *this one* |
| 2 | #272 | the slide list steps aside once you've picked a slide |
| 3 | #274 | the toolbar is clipped at 320px, taking ⋯ (23 actions) off-screen |
| 4 | #278 | pinch to zoom the canvas |
| 5 | #275 | a right-click context menu (the editor had none at all) |
| 6 | #276 | press-and-hold opens that menu on a phone — **stacked on #275** |
| 7 | #273 | overlays dismiss on an outside tap — **stacked on #272** |
| 8 | #292 | a formatting bar for a text selection, with real lists and headings |

#273 and #276 contain their parent's commit as their base, so merging the parent first reduces each to its own single commit.

Only #275/#276 and #292 touch shared files (`editor.ts`, `render.ts`, `preview.ts`, the catalogs); the rest are one or two files each. All eight report `mergeable: true` against current `main`, and none carry a `CHANGELOG.md` hunk — that file conflicted on every one of them twice over, so each entry lives in a collapsed section in its PR body for you to place at release time instead.

</details>

No rush at all, and I'm happy to split, reshape or drop any of them — I know `CONTRIBUTING.md` asks for an issue first on anything substantial, which #292 in particular is.

PR #272, opened 2026-08-10
slides: the slide list steps aside once you have picked a slide

iamgeo92:
## The problem

Below 700px the side panels are not columns — `styles.css` deliberately turns
them into **overlay drawers** laid across the canvas:

```css
/* Panels become OVERLAY DRAWERS. As flex siblings the 188px sidebar left a
   214px canvas on a 402px screen — the thing you are editing became the
   smallest thing on screen. */
.ed-sidebar, .ed-props { position: absolute; top: 0; bottom: 0; z-index: 40; … }
```

That is the right call. But the thumbnail click handler was written for the
column layout:

```ts
item.addEventListener('click', () => this.store.goTo(i))
```

So on a phone, tapping a slide navigated correctly and then left the drawer
sitting **on top of the slide it had just navigated to**. Every slide change
cost a second trip to the ☰ toggle in the topbar just to see what you picked.

## The fix

The drawer steps aside when you pick from it.

Closing is scoped two ways, so nothing changes where the panel is not a drawer:

- **By width** — only when the panel is actually an overlay (`≤700px`, the same
  threshold the CSS and `applyPhoneChrome` already use). On a wide screen the
  list is a real column beside the canvas and stays exactly where it is.
- **By target** — only the thumbnail itself. The per-thumbnail **Duplicate** and
  **Delete** buttons already `stopPropagation`, so editing the list never
  dismisses it. Those are edits to the list, not a choice of slide.

## Before / after

Same gestures both sides: open ☰, tap slide 3.

slide drawer, before and after

| before — the list covers the slide you just chose | after — the drawer steps aside |
|---|---|
|  |  |

In the "before" clip you can see the extra step the old behaviour forced: find
☰ again, press it, *then* look at your slide.

## How this was verified

Dev server, Chrome DevTools iPhone emulation (390×664, DPR 3, `hasTouch`) with
real touch dispatch, plus a 1280×800 mouse context for the desktop cases:

| case | result |
|---|---|
| drawer starts closed on a phone | pass |
| ☰ opens the drawer | pass |
| picking a slide closes the drawer | pass |
| the deck navigated to the slide that was tapped (not just closed) | pass |
| Duplicate on a thumbnail adds a slide **and leaves the drawer open** | pass |
| desktop: sidebar is a column and starts open | pass |
| desktop: picking a slide leaves it open | pass |

Gates:

```
```

**Scope note:** this PR deliberately covers only *picking a slide*. Dismissing a
drawer by tapping the canvas outside it is a separate gap and a separate PR, so
the two do not tangle in review.

**Caveat:** verified under Chromium's iPhone emulation rather than on physical
iOS hardware. This change is layout-threshold and event-target logic with no
WebKit-specific behaviour in it, so the risk of an emulation-only result is low.

---

<details>

```markdown
- **Picking a slide on a phone now shows you the slide.** Below 700px the slide
  list is not a column beside the canvas — it is a drawer laid over it. Tapping
  a thumbnail navigated correctly and then left the list sitting on top of the
  answer, so every slide change cost a second trip to the ☰ toggle to see what
  you had chosen. The drawer now steps aside when you pick from it. On a wide
  screen the list is a real column and rightly stays where it is, and the
  per-thumbnail Duplicate and Delete buttons leave it open on every width —
  those are edits to the list, not a choice of slide.
```

</details>


iamgeo92:
@nyblnet — review whenever suits you.

Small one: below 700px the slide list is an overlay drawer (which is the right call — the CSS comment explains why), but tapping a thumbnail left it sitting on top of the slide it had just navigated to, so every slide change cost a second trip to ☰.

Scoped two ways so nothing changes elsewhere: only while the panel *is* a drawer, and only from the thumbnail itself — the per-thumb Duplicate/Delete already stop propagation, so editing the list never dismisses it.

One file, +25/−1. #273 is stacked on this one.

PR #279, opened 2026-08-10
CI: Add PR build-size reports

7jameslondon:
## Summary

Related discussion: https://github.com/nyblnet/bento/discussions/178

This PR adds a CI workflow that will leave a comment on PRs that shows how the PR will change the build size. The workflow follows GitHub's privilege-separation recommendation. There is an [example on my fork of bento](https://github.com/7jameslondon/bento/pull/5).

## Validation
Verified on my fork with fork PRs to make sure everything works as it should.


nyblnet:
Merged — thank you, this was carefully built.

The part I want to acknowledge specifically is the privilege separation, because
it is the thing most implementations of this get wrong. The job that runs PR
code holds no token, the job holding the token never checks out or executes PR
code, and the artifact crossing between them is treated as hostile: name checked
and size-capped before download, exactly one expected filename, schema-checked,
and passed to the script through `env:` rather than interpolated into shell or JS
source. Actions pinned to full SHAs throughout. I reviewed all of that by hand
before merging and found nothing to change.

One change on our side: `PLAN.md` moved to `docs/pr-build-size-design.md`. Pure
rename, no edits — `docs/` is where this project keeps its reasoning, and
`-design.md` is the suffix the existing specs use (`collab-design.md`,
`relay-design.md`).

Worth saying why it was kept rather than dropped, since I nearly argued for
dropping it. The Security boundary section is the written form of why this is
two workflows instead of one, including "Do not use `pull_request_target` to
build the PR" stated outright. The split reads as ceremony to anyone meeting it
cold, and collapsing it back is the classic repository-takeover bug — that
document is what stops the edit. It earns its place.

Note there is an existing `Spaces shell size` step in ci.yml
(`scripts/test-spaces-size.mjs`) which is enforcing rather than advisory and
covers one of the three shells, so there is some overlap to reconcile later.
That is ours to sort out, not something to fix here.

PR #292, opened 2026-08-10
slides: format text by selecting it, with real lists and headings

iamgeo92:
## The problem

A text box already **supported** bold, italic and underline. It just never
offered them. The only two routes were `⌘B`/`⌘I`/`⌘U` and typing markdown —
neither of which the interface mentions anywhere except one hint buried in the
Typography panel. And lists and headings could not be made at all: everything
you typed came out as one flat run.

```
before:  Quarterly review<div>Revenue up 24%</div><div>Churn down to 1.8%</div>…
after:   <h1>Quarterly review</h1><div><ul><li>Revenue <b>up 24%</b></li>…</ul></div>
```

## What this adds

Select words inside a box you're editing and a bar appears over them:

**B** · *I* · <u>U</u> · ~~S~~ · `</>` │ **Title text** · **Heading** · ¶ │ • list · 1. list │ clear formatting

It follows the selection, reflects what is already applied (the pressed states
come from `queryCommandState`), and disappears when the selection collapses to a
caret — a bar over a caret has nothing to act on and would just cover the words
being typed.

Lists and headings are **real `ul`/`ol`/`li` and `h1`/`h2` in the document**, so
one renderer means the canvas, sidebar thumbnails, the presenter and PDF export
all agree. Heading sizes are in `em`, so a box keeps its hierarchy in proportion
when it's resized.

rich text, before and after

| before — one flat run | after — title, list, bold span |
|---|---|
|  |  |

## Why semantic tags rather than styled spans

The sanitizer strips **every** attribute unconditionally. That's the format's
whole defence, and I did not want to weaken it, so every tag added to the
allowlist is attribute-free by construction:

```ts
'UL', 'OL', 'LI', 'H1', 'H2'
```

A `<span style="…">` route would have required the sanitizer to start reasoning
about style declarations, which is exactly the thing `docs/PLATFORM.md` and the
untrusted-input work are built to avoid. The cost is that per-span **colour and
size** aren't offered here — those remain element-level properties. That felt
like the right trade rather than a gap worth opening the sanitizer for.

## Three things this had to get right

**1. `⌘B` could silently do nothing — a pre-existing bug this fixes.** Browsers
may express bold as `<span style="font-weight:bold">` instead of `<b>`; the
sanitizer strips the attribute, so the formatting was thrown away on commit and
the text came back unchanged. Caught by a test, not by reading — the first run
returned:

```
"alpha <span style=\"font-weight: normal;\">beta</span> gamma"
```

`execFormat()` now asks for the tag form explicitly, and **both** the bar and the
existing `⌘B`/`⌘I`/`⌘U` handler go through it so they cannot drift.

One case genuinely *cannot* be represented: you cannot turn bold **off** inside a
box whose own `fontWeight` is already bold, because no attribute-free tag means
"lighter than the box". Rather than a button that quietly does nothing, it is
disabled there and says why. Removing a `<b>` the selection sits inside *is*
representable, so that case stays enabled.

**2. File-manager thumbnails.** `preview.ts` inlines the runtime CSS **by hand**
for the thumbnailer, which runs no script and never inflates the compressed
stylesheet. Without list and heading rules added there, a deck's thumbnail would
render bullets as unmarked, unindented lines — a thumbnail misrepresenting the
slide it stands for. `node scripts/test-preview.ts` passes.

**3. List markers on centred text.** An `outside` marker is pinned to the box's
start edge, so on centred or right-aligned text it detaches from the line it
belongs to and floats off on its own — visible in my first screenshot. `render.ts`
now publishes the element's alignment and the marker moves `inside` for those
cases, in both the stylesheet and the preview inliner.

## Compatibility

The format stays additive. A shell that predates a tag **unwraps it and keeps the
words**, so a deck using lists opens everywhere — it just loses the bullets until
the reader updates. Sanitizing runs at *render* time, not load or save, so that
degradation is visual only: an old shell opening and re-saving a deck preserves
the markup unless that particular box is edited there. `docs/format.md` is
updated to state the new subset and this behaviour.

## How this was verified

Dev server at 1280×820, plus an iPhone profile (390×664, `hasTouch`):

| case | result |
|---|---|
| a selection raises the bar; a bare caret does not | pass |
| **only the selected span** becomes bold (`alpha <b>beta</b> gamma`) | pass |
| the pressed state reflects what the caret is inside | pass |
| a bulleted list is real `<ul>` with one `<li>` per line | pass |
| a numbered list is real `<ol>` | pass |
| a heading is a real `<h1>` | pass |
| **it survives sanitize + commit into the document model** | pass |
| …and re-renders as a real list on the canvas | pass |
| the bar comes down when the edit commits | pass |
| phone: the bar appears, fits the screen, targets ≥36px | pass |
| phone: a list can be made entirely by touch | pass |

```
npm run build:single                                            ok (668KB shell)
node scripts/test-preview.ts                                    18/18
node scripts/build-i18n.mjs --check                             current
```

## i18n

11 new strings in **all eight** catalogs, `packed.ts` regenerated. `Bold` reuses
the existing key (same sense — the font-weight label). `Title` deliberately does
**not**: it already means the *deck's* title in the About dialog, and
`docs/DECISIONS.md` — *"One English word, two meanings = two keys"* — says the
more specific site takes a qualified key, so the text style is `Title text`.

## Process note

`CONTRIBUTING.md` asks for an issue or Discussion before anything substantial,
and this qualifies — it touches what a `text` element's `html` may contain. I've
written it as a PR because it's easier to judge a concrete diff than a
description, but I'm happy to move the discussion to an issue, split it (the
`execFormat` fix stands alone), or drop the block-level half and keep only the
inline bar if you'd rather approach lists differently.

Related: `markdown.ts` said *"slides have no headings/lists DOM"*. That comment
is now stale, and I've updated it rather than deleted it — typing `"- "` still
produces a `"• "` glyph on purpose, because autoformat fires mid-keystroke and
silently converting the line you're typing into a list item would move the caret
and capture Enter from then on. The bar is where a list is asked for explicitly.

**Caveat:** `execCommand` is deprecated. It is also the only API that edits a
contenteditable's DOM *and* the browser's native undo stack together, which the
editor already depends on for `⌘B`. A hand-rolled Range implementation that
didn't feed undo would be a worse answer, not a more modern one — but if you'd
prefer to go that way, this is the natural place to draw the line.

<sub><b>Changelog placement:</b> appended at the end of <code>[Unreleased]</code> rather than the top, so it merges without a hand-resolve — same as my other open PRs.</sub>

---

<details>

```markdown
- **Format text by selecting it — and real bulleted lists, numbered lists and
  headings.** A text box already supported bold, italic and underline, but the
  only ways to reach them were ⌘B/⌘I/⌘U and typing markdown — neither of which
  the interface ever mentions — and lists and headings could not be made at all.
  Select words inside a box you are editing and a small bar appears over them
  with bold, italic, underline, strikethrough, code, Title text, Heading, body
  text, bulleted list, numbered list and clear formatting. It follows the
  selection, shows what is already applied, and works the same on a phone,
  where it is the only route to any of this.

  Lists and headings are real `ul`/`ol`/`li` and `h1`/`h2` in the document, so
  they render identically on the canvas, in thumbnails, in the presenter and in
  a PDF. Heading sizes are relative to the box's own font size, so resizing a
  box keeps its hierarchy in proportion. A deck using them opens in an older
  Bento too — a shell that has not heard of a tag unwraps it and keeps the
  words, losing the bullets until the reader updates.

  Also fixed on the way: **⌘B could silently do nothing.** Browsers may apply
  bold as `<span style="font-weight:bold">` rather than `<b>`, and the document
  sanitizer strips every attribute — so the styling was thrown away on commit
  and the text came back unchanged. Formatting is now always asked for in tag
  form. One case genuinely cannot be represented and now says so instead of
  failing quietly: bold cannot be turned *off* inside a box whose own weight is
  already bold, so the button is disabled there and explains why.
```

</details>


iamgeo92:
@nyblnet — this is the one I'd genuinely like your steer on before it goes anywhere, please.

`CONTRIBUTING.md` asks for an issue or Discussion before anything substantial, and this qualifies — it changes what a `text` element's `html` may contain. I wrote it as a PR because a concrete diff is easier to judge than a description, but I'm equally happy to move it to a Discussion, split it, or drop the block-level half.

The shaping constraint was yours: the sanitizer strips **every** attribute unconditionally, and I didn't want to weaken that. So formatting is semantic, attribute-free tags (`ul/ol/li`, `h1/h2`) rather than styled spans — which does mean per-span colour and size aren't offered here.

Three things you'll want to look at:

- **`⌘B` could already silently do nothing.** Browsers may emit `<span style="font-weight:bold">` rather than `<b>`; the sanitizer strips the attribute and the formatting is gone on commit. A test caught it. Both the bar and the existing keyboard path now go through one `execFormat()`.
- **`preview.ts`** inlines runtime CSS by hand for the JS-off thumbnailer, so lists needed rules there too — otherwise every deck using them thumbnails wrong.
- **`markdown.ts`'s "slides have no headings/lists DOM"** comment becomes stale. I updated rather than deleted it: typing `"- "` still makes a `"• "` glyph on purpose, because autoformat fires mid-keystroke and silently turning the line you're typing into a list item would move the caret and capture Enter.

Also: `Title` was already the deck's title in the About dialog, so per `docs/DECISIONS.md` — *"One English word, two meanings = two keys"* — the text style takes the qualified `Title text`.

`execCommand` is deprecated and I've explained in the body why I still think it's the right call here; if you'd rather not take that dependency further, this is the natural place to say so.

bouzidanas:
I like this idea, but there are a few things that I think needs tweaking. 

1. the toolbar is too wide/big. I think it should collapse to maybe 2-3 buttons plus a more (...) button that shows a drop down.
2. It's too in the way imo. I see the usefulness on mobile but there must be an alternative location near to the element that doesn't put it where it covers a significant portion of the slide.
3. Should not be always on for all users. Needs to be togglable incase users don't want to use it.

These issues are mainly UX centric.

PR #293, opened 2026-08-10
feat(slides): Live broadcast

niemes:
## Live slide broadcast - present any deck online, no accounts for viewers

This implements live broadcasting for bento/slides: the presenter's current slide syncs to any number of viewers in near-real-time over the existing collab relay. Viewers need no account, no write access, no collaboration keys. The broadcast copy of the file is the whole capability.

### What it adds

- **Share menu → "Save broadcast copy…"** — exports a standalone viewer file of the current deck. It boots into a locked present-follow mode and runs the real renderer on its own copy of the document, so viewers see the same slides, transitions, and morphs as the presenter.
- **Speaker view → Broadcast toggle** — arms the broadcast (off by default, every show), shows the live viewer count, and syncs **navigation, laser pointer, and black screen**.
- **The relay stays blind** — the nav channel carries only a slide number (plus laser/black control), never content. The relay never holds a document.
- **No secrets in files or URLs** — the room name and the connect token are both derived (sha256) from the presenter's signing key, so a broadcast copy carries no keys and a shareable link carries no token.
- **Write is impossible** — control frames are ECDSA-signed and verified by the relay (trust-on-first-use key pinning per broadcast room); broadcast copies carry no keys at all.
- **The format stays additive** — `collab.broadcast` is an optional field; existing files open unchanged.

### The two sharing cases

**Not hosted — the file *is* the link.** A deck without a hosting URL exports a pure snapshot copy embedding only the room name and relay. Give the file to viewers: when you start presenting and arm the broadcast, every open copy connects straight to the room derived from your signing key and follows — nothing to share, no URL involved.

**Hosted — a URL instead of the file.** A deck with a hosting URL (Document properties → `meta.hostClient`) exports a copy that additionally carries reader credentials and joins the deck's collab room as a live reader replica, so slide content updates in real time as the deck is edited mid-show. The speaker view shows a full viewer URL (`host?room=<name>`) with a Copy button, and the same hosted copy can be re-pointed at any presenter's room via `?room=` — a replacement presenter takes over without re-exporting files.

### Laser pointer

The presenter laser got some work in this pass: 

- **Draws only while the mouse is held down** -> moving the pointer without a click never draws; a stroke starts on pointerdown over the slide and ends with a clean `off` frame on pointerup. For performance with the sync worker and it is also a more natural behavior.
- **A longer, smoother trail** -> an ~0.8s fading sweep built from tapered quadratic segments (halo + core), so the tail reads as a sweep rather than a smear.
- **No text selection while drawing** -> with the laser armed, dragging on the slide can never select text (`user-select: none` on the slide subtree). Text selection is not broadcasted, so endup as visual noise for the speaker.
- **Synced to viewers** -> stroke points travel as slide-fraction coordinates (~30fps), so the audience sees the same dot and trail and including a fix for ghost strokes between mouse releases.
- Added an svg based icon, because  U+1F7D2 char dont render on macOS.

### Relay

`server/sync-worker` gains broadcast control-frame handling (nav/laser/black with signature verification, per-room TOFU key pinning, `lastNav`/`lastBlack` replay for late joiners, presence counts). 
The relay must be deployed (`wrangler deploy`).

### Verification

- `node scripts/test-nav.mjs` — 15/15: unsigned, forged, and member-key frames dropped; valid frames fanned; `lastNav` replayed ahead of live frames; presence rises on connect and falls on close; rate limiter intact; laser/black security.
- `tsc -b` clean, `npm run build:single` OK; no CRDT changes (`scripts/test-sync.ts` untouched).
- Manual two-tab session: the copy follows slide-for-slide with transitions, a mid-show joiner lands on the current slide, disarm quiets the feed, stale copies clamp instead of crashing.
- New UI strings in all 8 locale catalogs.

**Checklist**
- [X] `npm run build:single` succeeds (from `slides/`)
- [X] Ran `node scripts/test-sync.ts` if I touched `slides/src/sync/`
- [X] New UI strings added to every catalog in `slides/src/i18n/`
- [X] Document format changes are additive and backward-compatible
- [X] Did not bump the version or cut a release (maintainers sign releases)


niemes:
Small note for local testing, the sync worker variable `DEFAULT_SYNC_HOST` needs to be changed, its hardcoded in online.ts (maybe we could use an env variable, would be easier to config for people self hosting the worker)

nyblnet:
Brought this up to `main` (119 commits) and got it green — pushed to your branch as three commits on top of yours, authorship intact. Three of the changes are more than a rebase and you should look at them; I'd rather you push back than find out at merge time.

**1. The transport moved to `slides/src/broadcast.ts`.** #313 moved the sync transport to `kernel/src/sync/online.ts` and left the slides file as a ten-line facade three weeks after you branched, so your 291 lines had nowhere to land. It's *new* code, not edits to the collab transport, and it's presentation machinery, so it lives in slides now and imports the two kernel primitives it uses (`signText`, `syncHost`). The protocol is unchanged. It carries its own `b64u`/`mintKeypair` rather than borrowing the kernel's `deviceIdentity`, since the broadcast signer is deliberately a different identity from the collab member key.

**2. `collab.broadcast` became `doc.broadcast`.** This is the one I'd most like your view on. Since #313 the kernel types `collab` with `room` and `key` **required** — minted complete, always. Your non-hosted copy carried `{ on:false, broadcast }`, the first document shape where `collab` exists without a room, and every kernel boundary failed to typecheck. Moving the field to the document keeps `collab`'s shape: a plain broadcast copy has *no* `collab`, a hosted one is an ordinary reader replica plus `doc.broadcast`. Behaviour is identical (no `collab` → never joins; `doc.broadcast` → boots the viewer; old shells open a plain snapshot either way). The design docs and your DECISIONS entry now say `doc.broadcast`. The alternative was casting past the type system at every kernel boundary, which I didn't want to hand you.

**3. The export-secrets failures — the rig was right both times.** `saveBroadcastCopy` now strips through `stripCollabSecrets()`, the one function every export uses, so a secret field added tomorrow is stripped here tomorrow rather than when someone remembers a hand-written list. And it writes through `serializeAuto`. I know you chose plaintext deliberately — *"a password-gated copy is useless to viewers"* — but silently writing plaintext from an encrypted deck is the exact leak class that rig guards, and a presenter who wants an open copy can drop the password first, in the open. 53/57 → 59/59.

**Two things the move exposed that were true on your original branch:**

- **The broadcast socket bypassed the offline switch.** `new WebSocket()` was invisible only because `online.ts` was the allowlisted transport file at the time; a viewer who'd switched on "nothing leaves this computer" would still have connected. It goes through `netWebSocket()` now, the one chokepoint. Offline rig 33/33.
- **A re-rendered slide nested a `<section>` inside a `<section>`.** `section.replaceChildren(buildSection(slide))` — Reveal reads that as a *vertical* slide, so the next arrow would descend into it instead of advancing. It replaces with the new section's `childNodes` now. Related: `buildSection` takes the loop index, because `main` added a `morphNext` lookahead that needs it.

**Smaller:** the guard that stopped a broadcast copy minting a live session had been inside `startSharing()`, now the kernel's; it sits at slides' two call sites instead. `test-nav.mjs` needs a live relay and CI runs none, and `test-ci-registered` refuses exemptions by design, so it's `server/sync-worker/nav-check.mjs` now — the relay's integration check, next to the relay, with run instructions in the header. Model key table regenerated.

**What's still ahead of a merge, and not mine:** the relay change (`server/sync-worker`) is ops' zone and must deploy *before* any client that depends on the control channel — the sequencing that bit this project on the keepalive change. And security should review the control channel and TOFU pinning; I've read it and found the design sound, but that's their call to make. Provenance clearance is the maintainer's.

This is the most ambitious PR in the queue and a design worth having. Thanks for the design docs — they're why a five-week rebase was tractable at all.

nyblnet:
Kernel-zone view on the two shape decisions that are mine. Nothing under `kernel/` is touched, so this is outside the serialization lock; the maintainer decides.

## Decision 1 — `doc.broadcast`, not `collab.broadcast`: agree, and it is the only right answer

`CollabCreds` requires `room` and `key` because the kernel's sync layer is *typed on the presence of a session*. `SyncDoc.collab?` means "this document has a session or it does not"; every transport, `shareEligible()`, `stripCollabSecrets`, and the export-secrets rig all read it that way. A `collab` with optional room/key would be the first document shape where `collab` present ≠ session present, and the cost lands in exactly the places that must not be ambiguous: `shareEligible()` connects "if the doc ARRIVED carrying collab" — a viewer copy with `{on:false, broadcast}` *carries collab*. That is a dormancy question (PLATFORM §5, "a fresh copy must never phone home") answered by accident.

`doc.broadcast` keeps the kernel's invariant intact and makes the two cases structurally distinct: a plain broadcast copy has no collab; a hosted one is an ordinary reader replica plus `doc.broadcast`. Additive, optional, ignored by old shells. Correct on every axis. The model.ts comment already says this well — keep it.

## Decision 2 — `slides/src/broadcast.ts`: agree with the placement, disagree with one thing inside it

**Placement.** Broadcast is presentation machinery — nav, laser, black-screen are Reveal's vocabulary. Type has no present mode; spaces has none today. If "follow the presenter" ever becomes a platform feature it should move under the lock *then*, with a second consumer to shape it, rather than be lifted speculatively now. `slides/src/` is right for a single consumer, and it imports only kernel *exports* (`signText`, `syncHost`, `netWebSocket`, `lsGet`) — which is the boundary that matters.

**The copy.** `broadcast.ts` carries its own `b64u` and `mintKeypair`, and they are byte-identical to `kernel/src/sync/online.ts:27` and `:68`, which are module-private. The stated reason — the broadcast signer is a separate *identity* from the collab member key — is right, but a separate identity does not need a separate implementation; it needs a separate *call*. Two copies of key-minting drift, and the kernel's copy is the one the export-secrets rig and the relay's commitment scheme are built against. Ask: **export `mintKeypair` (and `b64u` if needed) from the kernel and delete the copies.** That is a two-line kernel change under the lock; I can cut it as its own small PR ahead of this one, or fold it into the kernel lock's next landing — maintainer's call on sequencing.

## One thing for security, from the mechanism side

`doc.broadcast.relay` puts a **relay origin in the document**. Collab deliberately does not (`CollabCreds` has no host; the build decides) — I flagged that gap on #423 yesterday. Broadcast is the first shape where opening a file connects to a host *the file names*. The offline switch still gates it (good — the rebase fixed the raw-WebSocket bypass), but a document that can direct a viewer's browser at an arbitrary WebSocket origin is a different trust posture from a document that names a room on the build's relay. Not mine to rule on; it should be a deliberate choice, and if it stands, it is also the precedent that decides the #423 follow-up.

nyblnet:
One correction to my account above, because I gave the wrong reason and the right one matters more.

I said the raw `WebSocket` went unreported "because `online.ts` was the allowlisted transport file at the time." That's false — the offline rig has no allowlist; its only chokepoint is `kernel/src/net.ts` and it scans every app's `src/`. The actual reasons: the check that catches a raw network primitive landed on 2026-08-16 (#305), five days *after* this branch was cut; and because the PR was conflicting, GitHub built no merge ref, so CI ran against the branch's own head with the branch's own August rigs. **A green tick on a conflicting PR is a report on the rigs as they were when you branched, not on `main`'s.** Nothing about the fix changes; the socket goes through `netWebSocket()` now and the rig passes 33/33 against today's `main`.

Also a routing correction from ops: `server/sync-worker/` is the **kernel** zone, so the relay half of this PR is kernel's and security's to shape, and deploying it is the maintainer's. Ops owns the CI side only.

nyblnet:
Kernel-zone addendum — the relay half. `server/sync-worker/` is kernel's (START-HERE:50), so this is the zone owner's view on shape; deploying it stays the maintainer's.

## Verification over TOFU: agree, and it is better than TOFU even at 60 bits

Confirmed in the diff. The worker already verifies signed rooms by recomputing the commitment — `'w' + sha256b64u(pub) === name` (worker.js, the `w` branch) — and pins a TOFU key only for non-`w` rooms because nobody had noticed the broadcast name is *also* a commitment. It is: `broadcastRoom(seed)` is `trunc10(b64u(sha256(seed)))`, re-hashed on the truncated name while it starts with `w`. Deterministic, unsalted, and the presenter socket already sends `?w=<pub>`. The relay can recompute that exact derivation from what it already receives. **Delete the TOFU branch and its pinned-key storage; verify.**

Why it is strictly better, not just simpler: the failure modes differ. With TOFU, an attacker who reaches the room first with a colliding key **locks out the real presenter** — pinned, later mismatch refused. With verification, a colliding key is merely a second valid owner: worst case, viewers see conflicting nav frames until the presenter re-arms. Lockout is the worse outcome and TOFU is the design that produces it.

**One precondition** for the recompute to hold: the `?w=` on the presenter socket must be the *same key that seeded the name*. The diff shows `broadcastRoom(creds.owner)` in one place and `broadcastRoom(signerPub)` in another — if the hosted path seeds from the collab owner key but the socket authenticates with the broadcast signer, verification fails for exactly the case it matters. Slides, please confirm it is one key end to end (I'd expect the broadcast signer, since a plain copy has no collab owner).

## The 60-bit question, answered as mechanism

Second-preimage on `trunc10` is 2^60 expected tries, and each try needs a *real* P-256 keypair the attacker can sign with — incremental point addition brings that to roughly a microsecond per candidate on CPU, faster on GPU. That is months of a large GPU farm, not an afternoon. For a control channel carrying a slide number and a laser dot, the realistic adversary is a prankster with the link, and 60 bits is far outside their reach. Two honest caveats: the truncation exists for short viewer URLs, and lengthening it is a one-constant change if security wants margin; and it should be *written down* as a chosen bound rather than left as an unexplained `.slice(0, 10)`.

## Deploy shape — the zone's rules, restated for this PR

- The relay change lands as **its own PR**, deployed **before** any client that depends on it; the client feature-detects and degrades to "broadcast unavailable" against an old relay. Slides has said it will take that client side — good.
- It must stay backward-compatible with every shipped client (PLATFORM §5), and it must **not** regress the Hibernation API — a `server.accept()` anywhere in the new socket path silently breaks all live collab on the free tier. The diff reads as using the hibernation handlers; the reviewer of the carved-out PR should check that explicitly.
- `test-relay-protocol.ts` must cover the new frame kinds and the verify path, including a socket whose `?w=` does not recompute to the room name being refused.

With TOFU gone, the diff shrinks to signed control frames, read-only sockets and the buffer — which is a change I'd be comfortable holding under the kernel lock when it is carved out.

PR #295, opened 2026-08-11
slides: restore mobile Save button outline and fix dark toast text contrast

bouzidanas:
## What & why

This PR fixes four small editor UI issues:

1. **Mobile Save button outline**

   Below 700px, the save-as dropdown caret element is hidden, but Save retained the styling making it still look like the left half of a split button. This PR restores the missing right border and corner radii so Save renders as a complete button.

2. **Toast text contrast**

   In dark mode, toast backgrounds switch through the theme while their fixed white text did not. Toast text now uses `var(--surface)` so it remains readable in both themes.

3. **Primary button text contrast**

   `.ed-btn-primary` uses `var(--ink)` as its background. Since `--ink` changes from dark in light mode to light in dark mode, a fixed white text color becomes unreadable in dark mode. Its text now uses `var(--surface)`.

4. **Active set-chip text contrast**

   `.ed-setchip.active` has the same inverted theme treatment and the same fixed-white-text issue. Its text now also uses `var(--surface)`.

### Mobile Save button

Mobile Save button outline before and after

### Toast

Dark-mode toast text contrast before and after

### Primary Share action

primary-share-action-text-contrast-in-dark-theme

### Active hover-set chip

active-hover-set-chip-in-the-real-editor-on-dark-theme

## Verification

- Typecheck passed
- `npm run build:single` passed
- Browser-tested the folded topbar at 390px
- Browser-tested the toast in light and dark modes
- Browser-tested the primary Share action in dark mode
- Browser-tested the active hover-set chip in dark mode

## Checklist

- [x] Read the relevant repository guidance before making changes
- [x] `npm run build:single` succeeds from `slides/`
- [x] No sync code was changed
- [x] No new UI strings were added
- [x] No document-format changes were made

### DIFF
--- a/slides/src/styles.css
+++ b/slides/src/styles.css
@@ -765,7 +765,7 @@ a.ed-btn { text-decoration: none; }  /* the update block's "What's new" link is 
   left: 50%;
   bottom: 26px;
   transform: translate(-50%, 12px);
-  background: var(--ink);
+  background: #1e2a3a;
   color: #fff;
   font-size: 13px;
   padding: 9px 16px;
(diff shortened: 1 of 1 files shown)


bouzidanas:
There is a similar issue (as the toast text color) with the button in the toast/info dialog that appears at the top when editor is opened in Safari browser in dark mode.

Edit: latest commit fixes this now.

PR #306, opened 2026-08-15
Export the current slide as an image, or all main slides as one ZIP

### LINKED ISSUE #261: Export slides as images (PNG/JPG) — current slide, or all slides as one ZIP
**Use case:** a Bento deck is perfect to hand to people as one file — but folks keep asking me for the slides as *images* too: Instagram carousels, printed handouts, sharing one slide in a chat. Right now the only way out of the app is PDF via print.

**Proposal:**
- Export the current slide as PNG/JPG, and "export all" as a single ZIP (`slide-01.jpg`…) so it's one save dialog.
- Rendered at 1.5–2× deck resolution. Feels like it belongs under **Save ▾**.

I actually built a working prototype — a small companion script (~300 lines vanilla JS, no deps) that clones the live `.bento-slide` into an SVG `foreignObject`, rasterizes via canvas, and zips with a tiny STORE-only writer. Works offline on `file://`. Happy to share it or open a PR — though honestly a native version rendering straight from the doc JSON would be much cleaner than my DOM-cloning, and would fit your UI properly.

A few gotchas I hit that any implementation will meet:
- collected CSS must be CDATA-wrapped inside the SVG (the editor's cursor styles embed raw `<svg>` in `url()` and break XML parsing);
- the SVG has to load from a `data:` URI — on `file://` a `blob:` URL taints the canvas;
- entrance-animation states need settling before capture.

(Same script also grew an **autoplay** toggle — advance every N sec + loop — great for kiosk/wedding slideshows. Can file separately if that's interesting.)

Love the project — the one-file idea is genuinely great. Thanks!

lazyeo:
Closes #261. This also implements the workflow requested in [discussion #243](https://github.com/nyblnet/bento/discussions/243).

**What & why**

Save ▾ now includes **Export slides as images…**:

- export the current slide as PNG or JPEG at 1× or 2×, including a current hidden or state slide;
- export all main slides as one STORE ZIP with ordered names (`slide-01.png`, `slide-02.png`, …);
- keep PNG transparency and use a white matte for JPEG;
- skip hidden and state slides in all-main export, matching presentation order.

The implementation uses a document snapshot and the shared renderer. Export is read-only except that an active text edit is committed after the user confirms the dialog. It does not regenerate `docId`, change `#bento-doc`, invoke save/update code, or overwrite the deck. It adds no feature dependencies.

The export remains self-contained and offline. Resource preflight accepts only bounded, supported embedded static images and fonts, refuses linked or unsupported resources before mounting a render, and re-checks the selected slide immediately before rasterization. Application CSS is restricted to explicitly app-owned stylesheets, while author CSS animation, transitions, and SMIL motion are frozen at their base frame for deterministic output. Dimension, payload, decoded-pixel, and archive budgets fail closed.

Password-protected decks show an explicit warning that exported images and ZIP files are plaintext; the original encrypted `.bento.html`, document identity, and model remain unchanged.

Tests are split by responsibility:

- pure contracts cover selection, naming, sizing, resource policy, image headers, budgets, ZIP bytes, and an independent ZIP reader;
- the browser rig covers representative pixels and security boundaries, zero export requests with a positive-control request, cancellation/atomicity, and real `file://` rasterization;
- shipped-file acceptance drives the built desktop/mobile UI and verifies current-slide PNG, ordered all-main ZIP output, and the encrypted-deck warning.

The compressed shipped file is 716,123 bytes versus 694,151 bytes for current upstream `main`, an increase of 21,972 bytes (3.2%) with the same feature dependencies.

Current browser measurements use Chrome. Safari/Firefox, the native OS save panel, and tray/WebExtension writeback are outside this PR's automated coverage.

**How I verified it**

- `npm run build:single` from `slides/`
- `node scripts/shell-gate.mjs slides/dist-single/Bento_Slides.bento.html`
- `scripts/test-slide-image-export.ts`: 204/204
- `scripts/test-slide-image-export-browser.ts`: 66/66 in Chrome 152 via `file://`
- `scripts/test-slide-image-export-acceptance.ts`: 23/23 against the shipped file
- `scripts/test-sanitize.ts`: 140/140 in Chrome
- `node scripts/test-preview.ts`: 28/28
- `node scripts/test-export-secrets.ts`: 57/57
- `node scripts/test-offline.ts`: 33/33
- `node scripts/test-sync.ts`: 45,368 checks
- slides and kernel typechecks
- i18n packed-table and coverage checks: 793 strings × 8 core locales

**Checklist**

- [x] Read the relevant parts of `CLAUDE.md` and the platform/collaboration docs
- [x] `npm run build:single` succeeds from `slides/`
- [x] The feature diff does not touch sync; the convergence rig still passes after the upstream merge
- [x] New UI strings are present in every core catalog and `packed.ts` is current
- [x] No document-format change; export preserves the existing document identity and does not serialize or overwrite the deck
- [x] Did not bump the version, sign, publish, or cut a release


nyblnet:
Thank you for this, and for the design. It is the one I followed. Export as images is on main now (0c92fe7): this slide or every slide in the show, PNG or JPEG at 1x or 2x, one file per page, hidden and state slides left out like print. I kept it smaller than this PR, about 6 KB in the shell, by leaning on the checks the app already has and by writing one file per slide into a folder instead of a ZIP; the rest of what you built here is the right shape for the converter app, which will own the heavy version later. It will be in the next release; it is not in a download yet. Closing this one with credit to you in the changelog.

PR #314, opened 2026-08-16
Clamp fraction digits so formatValue keeps its never-throws contract

eeshsaxena:
`formatValue` is documented "Never mutates, never throws", but it throws a `RangeError` on a format string with a long fractional part:

```js
formatValue(1234.5, { type: "number", format: "0." + "0".repeat(101) })
// RangeError: minimumFractionDigits value is out of range.
```

`readPattern` takes the decimal count straight from the pattern, so `0.` followed by enough zeros yields a fraction-digit count above 100, and `Intl.NumberFormat` rejects anything over 100. Since `format` is document data and this runs for every cell, one pathological pattern crashes the render. I clamp the count into Intl's accepted range; valid patterns (`#,##0.00`, `0.0%`, etc.) are unchanged.

### DIFF
--- a/dash/src/format.ts
+++ b/dash/src/format.ts
@@ -62,7 +62,10 @@ export function formatValue(v: unknown, col: Pick<Column, 'type' | 'format'>): s
   const p = readPattern(col.format)
   const isPct = p.pct || t === 'percent'
   const shown = isPct ? n * 100 : n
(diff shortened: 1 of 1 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Clamp fraction digits so formatValue keeps its never-throws contract`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining bento!

PR #315, opened 2026-08-16
tray/ios: find a document by something inside it

nyblnet:
Document search on iOS, matching what `tray/webext` has had and what `tray/android` is getting. Handover from the Android session; direction was already settled in `docs/DECISIONS.md` (2026-08-16).

`tray/webext` finds a deck by a phrase on a slide. `tray/ios` had the system browser's file-name search and contributed nothing to it — no CoreSpotlight, no `NSUserActivity`.

Two commits, split along the line that matters: the half that must **agree with the extension**, and the half that is **iOS's own problem**.

## 1. The extractor, and the thing that proves it agrees

`tray/ios/BentoIndex.swift` ports the extension's `extractText` — up to 40KB of prose per document, pulled out of the `#bento-doc` block as string values, data URIs stripped first, encrypted documents never read.

The decision entry allows each host its own copy on the grounds that a divergence here makes search *find less* — visible and recoverable — rather than corrupting a document the way a `bridge.js` divergence would. **That trade only holds if a divergence is loud**, so:

`scripts/test-tray-index.mjs` imports the **real `library.js`** — `describe()` is exported and takes injectable deps, so a fake file handle runs the shipping code path — and diffs it against the Swift, compiled on the fly with `swiftc`.

Diffing against the live reference rather than only against fixture files is deliberate. Static expectations pin a port to a *snapshot*: when the reference moves, both go stale together and the rig stays green while the two drift apart. Importing it means `library.js` cannot move without this failing.

**59 documents agree on every field** — 45 generated edge cases plus every document in the tree.

Finding the edge cases was most of the work. They sit where the two languages part company:

| | why it bites |
|---|---|
| **UTF-16** | every index, length and budget in the reference is UTF-16; the natural Swift spelling counts graphemes and disagrees on any document with an emoji or CJK in it |
| **`\s`** | JavaScript's includes U+FEFF and U+00A0, excludes U+200B; ICU's is `\p{Z}`-based |
| **`\d`** | JavaScript's is ASCII; ICU's is Unicode Nd, so `&#١٢٣;` would be eaten as an entity |
| **budgets** | 200-char data URIs, 400-unit values, 200-char tags — each tested at N-1/N/N+1 |

The scanners are hand-written, and each carries the argument for why one forward pass equals the reference's backtracking regex. It turns on `[^"\\]` being unable to match a quote, so no interior repetition boundary the engine could give back to can hold one — which means if the greedy run hits the cap without a quote, every shorter alternative fails too.

**One deviation.** The final `slice(0, 40KB)` can land inside a surrogate pair; JavaScript keeps the lone half and a Swift `String` cannot hold one, so it is dropped. Forced by the language, not chosen, and better than U+FFFD in a search index either way. The rig accepts exactly that shape — one extra unit, unpaired high surrogate, identical prefix — and nothing looser.

## 2. The iOS surface

**A whole-folder grant**, which the app did not previously take — it could open one file at a time and could not enumerate anything. Folder-mode `UIDocumentPickerViewController`, persisted as a security-scoped bookmark.

Each indexed document also gets its **own** bookmark, minted inside the folder's open scope. Not tidiness: a file in a granted folder is readable only while the *folder* is scoped, and an editing session outlives the walk by minutes.

**Two surfaces, because iOS has two.** `UIDocumentBrowserViewController` stays the root — it is not a list but Files, iCloud Drive, every File Provider on the device, drag-and-drop, rename in place, favourites and tags — so search sits one toolbar button *beside* it. The same prose goes to **CoreSpotlight**, which is what the gap statement actually asked for; results return through `scene(_:continue:)` and resolve via the index rather than by re-deriving a path that may have moved.

**Two rules enforced rather than described.** Putting document text into a system-wide index is a real reach beyond the app, so: an encrypted document is never read for text or preview, and revoking a folder deletes its documents from CoreSpotlight as well as from the local list.

## Verification

Rigs: `test-tray-index` (new), `test-tray-bridge`, `test-webext-library`, `test-preview` all pass.

End to end on the iPhone 17 simulator: grant a folder → walk → search a phrase that appears **only in slide prose** → open the result in place. Then again for a folder **outside the app container** (On My iPhone, served by a File Provider), which is the case the container does not exercise. Grants and index survive relaunch; clearing the cached index and re-walking exercises the bookmark minting.

**One thing I checked and did not fix.** An edit made after opening from search does not write back within 10s. I ran the control — same external file, opened through the browser's own Recents instead of search — and it behaves identically, so this is pre-existing tray behaviour and not introduced here. The write path is `bridge.js`/`EditorViewController`, shared with Android, so I have left it alone and flagged it rather than changing a shared file mid-handover. Worth a look on its own.

## For the Android port

- **Run the same rig.** It takes `--corpus <dir>`, so the shared fixture corpus is additive rather than a replacement. A second rig would be a second opinion about what the reference says.
- **Count UTF-16 code units.** Kotlin's `String` is UTF-16 already, so this is nearly free there — but `codePointAt`/grapheme APIs are the trap.
- **The surrogate deviation may not apply**, since Kotlin can hold an unpaired surrogate. If so, Android matches the reference exactly and only Swift deviates — worth confirming rather than assuming.

## Notes

- `docs/DECISIONS.md` gains an entry for what is settled here that another host could contradict: CoreSpotlight as the iOS surface, the live-reference diff, and the UTF-16 requirement.
- CI is Linux with no Swift toolchain, so the rig **skips there — loudly**. A Mac has to run it before either side changes. That is the same convention as `test-tray-bridge.ts`.
- `tray/README.md` parity rows updated for search, plus two things named rather than left to be discovered: no thumbnail in the results list yet, and the extraction inheriting style values (`sans-serif`, `#EAF4FF`) as "prose" — which is the extension's behaviour too, so narrowing it is a change to all three ports.


nyblnet:
**Cross-checked against the shared corpus** (`tray/fixtures/` on `tray-android-search`), pushed as a third commit.

10 of 11 cases agreed as written. That is a genuine cross-check rather than a formality: this port was written against `tray/webext`'s `library.js`, and the corpus was generated from an independently written reference (`tray/doc-index.mjs`). Budgets match too — `textBudget` 40960, `sniffBytes` 65536, `headBytes` 307200.

The eleventh was a **contract difference, and the corpus is right**. `library.js`'s `describe()` always returns a title, falling back to the file name, because its caller sniffs first so it never faces a non-document. A port that answers every question in one call *does* face one, and the corpus says it should answer null throughout. `BentoIndex.describe` now folds `isDocument` in and returns nils; `LibraryIndex` supplies the file-name fallback — the better seam anyway, since inventing a title during extraction reports one for a file that isn't a Bento document at all.

`scripts/test-tray-index.mjs` now runs **both** checks, because they fail for different reasons:

- **the shared corpus** proves all three hosts give the same frozen answers — including Kotlin, which nothing on the machine running the Swift rig can execute;
- **the live `library.js` diff** proves this port tracks the extension as it *moves*. A frozen answer key pins a port to a snapshot: when the reference changes, static expectations go stale silently and every host stays green while drifting.

The corpus section **skips rather than fails** when `tray/fixtures/` is absent, since it arrives with the Android branch — a rig that fails until an unrelated branch lands is a rig people learn to ignore. Verified both ways: skipping without it, 11/11 with it copied in.

nyblnet:
**Starter decks are no longer bundled on iOS** — confirmed with Andy, and mirrored from Android's `Releases.kt` (fourth commit). Same apps, same key, same order of checks, same refusal messages, so the two hosts cannot answer differently.

The 686KB shell is out of the bundle. The host now makes **one** network request, only when creating a document, only to the signed release channel — cached in Application Support (not Caches, which the OS evicts), so only the first "New" of a release needs a connection.

**Verification is why the fetch is acceptable**, and it is the part worth reviewing. Both halves, as `kernel/src/update.ts` does them: ECDSA P-256 over the *exact* payload string, then the downloaded shell's sha256 against the hash pinned inside that signed payload. `tray/webext`'s `newDocument()` does neither — and is separately broken against the live server, reading a top-level `url` from what is actually a `{payload, sig}` envelope. It was not copied.

**A verifier that has only been seen saying yes is `return true`.** `scripts/test-tray-releases.mjs` runs the Swift verifier against a captured real manifest plus seven tampered ones:

| case | expected |
|---|---|
| the real manifest | ACCEPT |
| payload edited to swap the download url | REFUSE |
| payload edited to swap the pinned hash | REFUSE |
| signature bent | REFUSE |
| a genuine signature over DIFFERENT bytes | REFUSE |
| signature removed | REFUSE |
| signature empty | REFUSE |
| captive-portal HTML instead of JSON | REFUSE |

8/8 behave correctly. The fixture does not go stale — a signature stays valid for the bytes it covers forever.

Two details inherited from Android rather than rediscovered: the signature is raw `r‖s` not DER (CryptoKit takes it directly, so iOS skips the conversion Java forced there), and `Accept: */*` — a browser-shaped Accept got the artifact treated as a page being browsed and a CDN injected a beacon into it, 359 bytes longer than the signed release, which the hash check correctly refused.

Verified on the simulator: "+" offers the three apps; Slides fetches, verifies, and opens, and the created file's sha256 equals the manifest's exactly (`9fef088b…`, v1.0.18); the second creation is served from cache; Spaces answers "Spaces has not been released yet" from the 404 rather than an HTTP code, leaving no empty document behind.

One bug found and fixed while testing: the editor did not open after creation, because presenting it raced the progress alert's dismissal — the same class of trap the surrounding code already documents for the browser's own transition. The dismissal is now awaited rather than assumed.

PR #329, opened 2026-08-22
spaces: a card for a link, filled in by hand because nothing is fetched

nyblnet:
## What changed

A `link` block type for bento/spaces — a card for an address on the web, the
outward-facing sibling of the existing `pagelink` page card. It carries `url`,
`title`, `desc`, `site`, `icon` (an emoji) and an optional `image`, and **every
field is stored in the file**.

Insert it from the `/` menu or the Insert dropdown ("Link to the web"). Editing
is the card's own hover button, plus an entry in the block menu — which is also
the touch sheet, so a card made on a phone can be changed on one.

## The decision: nothing is fetched, and not just at render time

The render-time half was already settled: PLATFORM §1 (no network to open,
edit, read or save) and "A space does not phone home when it is opened"
(DECISIONS, 2026-08-03). A card that fetched on open would be the tracking
pixel that decision was written about.

The interesting half is the **editor**, where `about.ts`'s update check is
precedent that an explicit, documented network touch can be on-platform. It
still does not belong here, for three reasons in order of weight:

1. **It cannot work.** Reading a url's OpenGraph tags means reading a
   cross-origin HTML body — opaque under `no-cors`, a CORS failure otherwise,
   and an outright failure from `file://`, the origin this format exists for.
   Notion and Slack do this on a server. Shipping it means running a proxy,
   i.e. a backend, i.e. off-platform.
2. **The update check is a different shape**: one origin we control, opt-out,
   verified against a key already in the file. A metadata fetch contacts an
   *arbitrary third party named by the document*.
3. A stored card is legible in the JSON, diffable, works offline, and cannot
   rot into a request when someone later adds a refresh button.

So the honest v1 is the card the author fills in, and the dialog says so in the
first line: *"Nothing is fetched. A card shows what you type here — opening this
space never contacts the site."*

Full reasoning appended to `docs/DECISIONS.md`.

## The model

```jsonc
{ "id": "b7", "type": "link",
  "url": "https://example.com/docs",   // https:, http: or mailto: only
  "title": "The guide",                // absent falls back to the url
  "desc": "Everything about the thing",
  "site": "example.com",               // absent is derived from the host
  "icon": "📘",
  "image": "asset:<key>",              // asset:/data: only; a remote one is DROPPED
  "html": "<a href=\"https://example.com/docs\">The guide</a> — Everything about the thing" }
```

`model.linkCard(b)` resolves a card once for every surface — pure, DOM-free,
and therefore testable in node, which is where the negative test lives.

**Untrusted input.** The url is a block FIELD, so it never passes through
`sanitizeInline`. `sanitize.ts` now exports `externalHref()`, the outward half
of `HREF_OK`, tested against the **raw** string for the reason already
documented there. Allowlist, not blocklist: the URL parser strips tabs,
newlines and leading whitespace from a scheme, so ` javascript:` and
`ja\tjavascript:` are both `javascript:` when followed, and neither starts with
`https:`. A url outside the list renders as a **dead card that keeps its
title** — it degrades, it never becomes an `<a>` to nowhere.

**No remote images.** `image` is `asset:` or `data:`; a remote one is dropped by
`linkCard()` rather than deferred behind a "load this image" button. An image
block's placeholder stands in for the block's whole content, so there is
something to consent to; a card's thumbnail is decoration beside a link that
already works. The editor's picker runs the same `prepareImage` → `internAsset`
path an image block does.

**Additivity.** `html` is written beside the fields by ONE writer
(`editor.applyLinkCard`), the same contract `prop` keeps — so a build predating
this type renders a working inline link, and an unknown-type file round-trips
untouched either way.

**`toMd`: a link card in Markdown is a link** — `[title](url) — description`.
Not a table, not an html `<div>`: every card-shaped export stops being a link
the moment it leaves this app, and where it points is the one fact that cannot
be reconstructed. Brackets in a title are escaped; a url holding a space takes
the angle form; an untitled card uses the url as its own label; a card with no
valid url exports as plain text, so a reader is never handed something to click
that goes nowhere.

## Deliberately not here

**Pasting a bare url as a card.** The paste pipeline is another agent's zone,
and a second document-level `paste` listener would race the first. No paste
handling of any kind is in this branch.

**Translated menu label.** `SPECS` labels are data, and the i18n sweep reads
literal `t()` calls out of the source — so "Link to the web" ships English in
all eight locales, exactly as every other block menu label already does
(`blocks.ts` documents this). Every UI string that *is* swept is translated in
all eight catalogs.

## Verified

- `node_modules/.bin/tsc -b` clean in `spaces/`.
- `scripts/test-spaces-model.ts` — **514/514** (was 460). New checks include:
  eleven hostile url forms that never become clickable; the empty-card and
  no-title fallbacks; **the negative** — six shapes of remote `image` dropped,
  `render.ts` carrying no `fetch`/`XMLHttpRequest`/`new Image()`, and the link
  case reading its only `img.src` from the already-filtered value; the two href
  allowlists cross-checked so the html fallback cannot be silently stripped by a
  stricter later build; and every `toMd` form.
- `test-spaces-agent.ts` 143/143 · `test-spaces-calc.ts` 90/90 ·
  `test-spaces-journal.ts` 45/45 · undo rig (esbuild-bundled as ci.yml does)
  18/18.
- `build-spaces-i18n.mjs --check` — 287 strings × 8 locales, all complete.
- `npm run build:single` + `node scripts/shell-gate.mjs` on the output — clean.
- Size: **166,482 → 171,074 B** against the unchanged **176,128** ceiling
  (97.1%). `test-spaces-size.mjs` passes; no budget raise.

### Browser evidence (Chrome, served copy, `location.pathname` asserted)

Built a uniquely-named copy (`/linkcards-proof-a7f3.html`, served on
127.0.0.1:5186 — the pane caches `file://`) whose document carries four cards:
a full one, one whose `image` is `https://tracker.invalid/pixel.png`, one whose
url is `javascript:alert(1)`, and one that is completely empty.

Measured at `location.pathname === "/linkcards-proof-a7f3.html"`:

| card | element | href | rendered |
|---|---|---|---|
| full | `A` | `https://example.com/docs/guide` | `target=_blank`, `rel=noopener noreferrer`, "📘example.com" |
| remote image | `A` | `https://example.org/paper` | **0 `` elements** |
| `javascript:` url | `DIV` | — | `sp-dead`, title "Do not click" kept |
| empty | `DIV` | — | `sp-dead`, "A link with nothing in it yet" + "Add a link" |

**The no-network proof:** `document.images` was empty for the whole page and
`performance.getEntriesByType('resource')` returned `[]`. `read_network_requests`
showed exactly one entry — `GET blob:http://127.0.0.1:5186/…` (the compressed
shell inflating its own module) — and nothing else. No request to
`tracker.invalid`, and none to anywhere.

Then exercised the insert path in the same tab: Insert → "Link to the web" →
the dialog (labels *Web address / Title / Description / Site name / Icon*, the
"nothing is fetched" note, and the picture buttons) → Save. The committed block
was `{type:'link', url, title, desc, icon, html:'<a href="…">Bento Spaces</a> —
…'}` — empty `site` omitted, host derived (`www.` stripped), `html` written
beside the fields — and the resource list was **still empty** afterwards.

§1 (no network — reinforced, with a test), §3 (format additivity — new open
block type + `html` fallback), §8 (i18n — all eight core catalogs). No
`kernel/` changes.


nyblnet:

PR #330, opened 2026-08-22
spaces: a remark belongs to a block, and never to the printed page

nyblnet:
Comments for bento/spaces: leave a remark on a block (Block options → Comment)
or on a whole page (the page's ⋯ → Comment on this page), reply to it, resolve
it. Threads are saved in the file so they travel with it, and they never appear
in the reading view or on paper — a comment is workspace, not document.

## The anchor model, and why

Two anchors and no third: a **block** (`Block.comments`) or a **page**
(`Page.comments`). A deck is a canvas, so slides anchors a comment to an
(x, y); a space is a tree of pages of blocks, and the block is the thing that
already has durable identity — ids are unique document-wide, never reused, and
links, backlinks and the CRDT all key on them. A text RANGE inside a block is
deliberately not an anchor (deferred): an offset pair has no meaning after the
concurrent edit that moved it, and the format is permanent.

The anchor is expressed as **storage**, not as a field, and that is a
collaboration decision. Under the shared engine every non-container property is
one last-writer-wins register per (node, key), so a single `Page.comments`
array — slides' shape transliterated — would make every thread on a page
contend for ONE register: two people commenting on two different paragraphs at
the same moment, and one comment is gone with nothing said. Per block, that
converges, because each block is its own CRDT node. It also settles two
lifecycle questions for free: deleting a block takes its threads with it
(undoably), and moving a block to another page carries them along.

## Collab caveat (documented, not pretended away)

Two people commenting on the **same** block concurrently, or on the same page,
is still last-writer-wins on that one array — the limitation slides' table
`rows` carries, for the same reason. Replies and resolve live in the array too,
so a reply racing another reply on one thread can lose. Making that converge
means threads as their own CRDT nodes: an engine change, and kernel changes are
serialized. Written down in `spaces/src/model.ts` beside the type and in
`docs/DECISIONS.md`.

## Where the UI lives

- **Not the gutter** — a block already has one at the start edge, holding two
  controls because a phone reserves 44px and fits one. A third affordance there
  is a control no thumb can reach.
- **Not a sidebar panel** — the sidebar is the page tree, the one navigation
  this app has.
- **The end margin**, opposite the gutter: outside the text column, so a marker
  never reflows the prose it is about, and beside the line it refers to. Page
  threads sit in a row under the title. The thread itself is the editor's
  existing popover (bottom sheet on a phone), so it dismisses like every other
  menu. The tree carries a **count** of unresolved threads — the one comment
  fact worth knowing from another page.

## Untrusted text

Comment text is **plain text** — not "sanitized like block html". Nothing is
ever parsed as html: the model stores a string, the UI writes it with
`textContent`. The rig asserts `innerHTML` appears nowhere in `comments.ts`.

## Agent surface

`bento.comments({ resolved?, pageId? })` → every thread, flat, each with a typed
anchor (`'block'` + `blockId`, or `'page'`). **Read-only, deliberately**: acting
on a remark is the agent's job, resolving it is the commenter's. Every field is
coerced to the type the report promises, so a hand-edited file cannot make it
throw halfway through. Documented in `docs/spaces-agents.md`.

## Format additivity

`comments` is absent in every existing file; a build predating this one
round-trips it untouched, including unknown fields **inside** a comment. Pinned
in `scripts/test-spaces-model.ts`.

## Verified

- `tsc -b` clean in `spaces/`
- model rig **478/478** (new: round-trip + unknown fields inside a comment,
  both anchors, hostile shapes, and the editor-only rule as source assertions)
- agent rig **155/155** (new: the verb, its filters, its coercions, and that
  reporting changes nothing)
- calc 90/90 · journal 45/45 · undo rig (esbuild bundle, as CI runs it) 18/18 ·
  `test-sync-spaces-session` 48/48
- `build-spaces-i18n.mjs --check`: 288 strings × 8 locales, all complete
  (18 new keys; 14 lifted word-for-word from the slides catalogs)
- `npm run build:single` + `scripts/shell-gate.mjs` clean; `test-spaces-size`
  ok, `test-i18n-coverage` ok
- **Real browser**, uniquely-named served copy (`/cm-a1.html`, asserted in
  every measurement): comment on a block and on a page, reply, resolve, badge
  appears and clears, badge visible from another page; marker measured outside
  the text column (block ends x=1124.5, marker at x=1137.5); a comment reading
  `plain <b>text</b> please` renders as those characters (`innerHTML` shows the
  escaped form); reading view paints 0 markers while the threads stay in the
  document; and the **actual print tree** built by the print path rendered 11
  pages with 0 markers and no occurrence of the substring at all. Undo takes a
  comment back and redo returns it; `serialize()` carries the thread in
  `#bento-doc`.

Screenshots stop partway through the session: the shared browser pane stopped
compositing (four sibling agents on the same browser), so the later checks are
DOM measurements taken in the same real browser rather than images.

## Size

166,482 → **170,878 B** (+4,396), inside the existing 176,128 B ceiling at
97.0%. `scripts/size-budgets.json` is **unchanged**.

Format additivity (§3) — new optional document data, gated by the model rig.
Nothing under `kernel/`.


nyblnet:

PR #331, opened 2026-08-22
spaces: a page can hold a clip, and nothing in it ever starts playing

nyblnet:
A `media` block for `bento/spaces` — ONE type carrying `kind: 'video' |
'audio'`, mirroring `image` down to the hybrid `src` (`asset:` · `data:` ·
`https:`), the remote-consent gate and the percentage width.

## The rule this is really about

**Autoplay is recorded and never obeyed, and that lives in a function.** Slides
learned the shape expensively: autoplay set at render time fires on the editing
canvas and in every thumbnail, so it belongs to present mode — the one surface
that owns playback. A space has no such surface. It has an editor, a reading
view, a printout and a file-manager still, and a clip that starts itself is
wrong in all four.

So it is not "the renderer happens not to set it", which is a property of one
function the next surface would have to rediscover. `mediaPlayback(b)` in
`blocks.ts` returns the flags a surface may apply, and its `autoplay` is typed
`false`. `Block.autoplay` still round-trips untouched (PLATFORM §3) — a build
that *does* own playback may legitimately write it, and there is no server to
migrate a file that already has.

## The rest

- **Paper and thumbnails get a still**, gated on `printing` — the flag that
  already means "this output cannot be interacted with". The poster frame if
  there is one, a labelled box if not; never nothing. `preview.ts` still bans
  `video,audio` outright, now as defence in depth.
- **A linked clip needs consent more than a linked image does**: a remote
  `<video>` asks its host for byte ranges the moment it is parsed, so opening
  the space *is* the ping. The preview passes no `allowRemote` at all, so a
  still built for a file manager can never make a request.
- **The URL box is its own gate.** `src` is not inline html and never reaches
  `sanitize.ts`, so "Use a link…" is an `^https?://` allowlist, never a
  `javascript:` blocklist.
- **`toMd` exports a link.** Markdown has no video; `` is image
  syntax and draws a broken-image glyph everywhere, and a bare URL is a player
  only on github.com. A link is correct in every renderer.
- **Nothing is re-encoded** — `MEDIA_EMBED_BUDGET` is 8MB (twice the image
  budget, matching slides), and the URL field is the escape hatch a "no" lands
  on, because a browser file picker hands over bytes and never a path.
- Fixed in passing: `validate()` swept `src` into its used-asset set but not
  `poster`, so a video's still frame would have read as an orphan.

## Verified in Chrome

On a uniquely-named served copy (`/media-check-a41.html`, asserted alongside
every measurement — the pane caches):

| what | result |
|---|---|
| `/` menu | offers "Video or audio · Plays in the page"; lands on the chooser |
| `javascript:alert(1)` in "Use a link…" | rejected, "That needs to be an http or https address" |
| linked clip, before consent | consent placeholder naming the host, 0 player elements, 0 requests for the clip |
| after "Load this video" | `<video controls preload=metadata playsinline>`, no `autoplay` attribute, `paused: true` |
| `play()` | reached t=0.60s of a 6s clip, then paused — the controls are real |
| 3 blocks, two declaring `autoplay: true` | all `paused: true, currentTime: 0` a second later, no `autoplay` attribute anywhere |
| reading view | same three paused; no editor chrome; `controls: false` honoured there while the editor keeps its own |
| print | 0 `<video>`/`<audio>` in `.sp-printroot`, 3 stills, 1 with the poster |
| static preview (`serializeAuto`) | 0 players, 0 `src` attributes of any kind; embedded clip as "▸ Audio" with its caption, linked ones as consent placeholders |
| saved `#bento-doc` round trip | `autoplay`, `loop`, `muted`, `controls`, `poster` all preserved, including the flag this build refuses to obey |

## Gates

`tsc` clean · spaces model **493/493** (was 464 — 29 new, pinning the model, the
autoplay rule three ways and the markdown form) · agent 143/143 · calc 90/90 ·
journal 45/45 · undo 18/18 through the esbuild bundle CI uses · i18n 296 × 8,
all complete · `shell-gate` pass.

## Size

166,482 → **172,722 B** (+6,240 B, +3.7%), same machine, same zopfli. The
ceiling moves 172 → 176 KiB in `scripts/size-budgets.json` with the numbers in
the note — the zopfli headroom spent on a feature, which is what it was left
for.

## i18n

26 new strings in all eight catalogs. Where slides had already translated the
same English — Video, Audio, Loop, Muted, Controls, Poster — it was lifted
rather than reworded; the rest follow each catalog's own terminology header
(pt stays European: "ficheiro", "controlos").

## Conflicts

`docs/DECISIONS.md` and `scripts/size-budgets.json` are shared with sibling
spaces work in flight; both are append/one-line changes.


nyblnet:

PR #332, opened 2026-08-22
spaces: a page can leave as its own space, and a space can arrive inside one

nyblnet:
The starter space says "a space is never a dead end: the whole document is
plain JSON in this file, and every page exports as Markdown." That was half a
promise — a page could not leave as a SPACE, and a second space could only
replace this one. M4's two exits finish it, both in one pure, DOM-free module
(`spaces/src/portable.ts`), so the round trip is asserted in node.

## What changed

**Export page as a space…** (save dropdown, ⋯ menu, About). `extractSpace()`
writes one page — and, if you want, its subtree — as a whole new `.bento.html`.
The dialog shows the real counts as you change the choice: how much travels,
how many links point out of the selection, how many images go.

**Import a space** — the existing importer, extended rather than duplicated.
The overlay gained an "Add pages under" picker that governs BOTH ways in
(Markdown and spaces), plus "Choose a space…"; dropping a `.bento.html` on the
window works too. `planGraft()` nests the arriving pages under the page you
chose.

## Decisions, and why

| Question | Answer |
|---|---|
| Identity of an extract | **Fresh `docId`, and `collab` removed entirely.** Keeping the id makes the extract a FORK — the session keys BroadcastChannel on `docId` and the saved file IS the capability, so three pages would join the two-hundred-page space's room. Keeping `collab` hands over `room`, `key`, `writerPriv`, `ownerPriv` and the invite chain: the capability to read and write the whole space. `template` goes too (it re-mints `docId` on open); `readonly` stays — upgrading someone's reading copy is not the exporter's call. |
| Links out of the extracted set | **Become the literal `[[Page title]]`**, the same thing an unresolvable wikilink becomes on the way in. Honest, searchable, valid Markdown, and it re-resolves if the halves are reunited. A silent unlink lies about the file; a live `#p/<id>` pointing at nothing looks fine until clicked — and after an import can land on a stranger page holding that id. Pagelink blocks likewise. |
| Which assets travel | **Only those the extracted pages reference.** Images are the only weight in a space; carrying all of them is a copy with pages hidden. Fonts are the exception and barely one — the theme names them, so every page uses them. |
| `doc.home` | Set to the extracted root. `homePage()` falls back to `pages[0]`, so a stale `home` would silently open the wrong page. |
| Id collisions on import | **Keep an id that is free; rename only a collision** — the rule `parseDoc` already applies inside one file, so links, backlinks and future CRDT keys survive an import that never had to touch them. Replacements come from `repairId` (now exported), derived FROM THE BYTES — never `Math.random`, never `docId`, which `template: true` moves. |
| Links inside the import | Repointed in the same pass, so nothing arrives pre-broken and no link lands on a host page that happens to hold that id. A link naming a page absent from the file becomes text. |
| Assets on import | **Content addressing does the work**: the same image already in the host merges to nothing. A key held with DIFFERENT bytes (impossible from the store, possible in a hand-written file) mints a `~n` variant, exactly as `internAsset` does — trusting the key alone would replace the host's image with the visitor's. |
| Undo | **One `store.commit`** → one ⌘Z, as the Markdown import is. |
| Trust | The file takes the ordinary load path and no side door: inert `DOMParser` read of `#bento-doc`, `parseDoc` refuses rather than degrades, an encrypted envelope is reported instead of guessed at, and `sanitizeInline` runs over every arriving block. |

Format additivity holds through both legs — unknown top-level fields, unknown
per-page fields and unknown block types round-trip untouched, asserted in both
directions. No kernel file was modified.

## Verification

- `npx tsc -b` clean in `spaces/`.
- **Model rig 499/499**, including the new round trip: extract a subtree, import
  it into a space that collides on every id and on an asset key, then assert no
  id collision, no dangling link, no credential in the bytes, the asset merge
  both ways, and identical ids across two planning runs.
- **Undo rig 25/25** (bundled, real `Store`): one ⌘Z removes the whole import
  and what is left is byte-identical to before; redo brings it back.
- Agent 143/143 · journal 45/45 (5 timezones) · calc 90/90 · spaces sync rig
  green · `build-spaces-i18n.mjs --check` complete in all eight catalogs (24 new
  strings, real translations).
- `npm run build:single` + `scripts/shell-gate.mjs` on the shell, **and on an
  exported file** — an extracted space satisfies the splice contract itself.
- **Real browser**, uniquely-named served copy (the pane caches). The pane
  denies File System Access, so the picker was stubbed to catch the written
  bytes. Exported the starter's Tracker subtree (6 pages, 31 blocks): the file
  passes the gate, opens as its own space, carries no `collab` in its bytes, and
  stays dormant — its only requests are the file and its own inflated blob, it
  mints fresh credentials as any new document does, and the live control reads
  "Not sharing yet". Dropped it into a SECOND starter where every id collides:
  37 ids renamed, no duplicate id, no dangling link, one ⌘Z restored the space
  byte-identically, and a second export carrying internal links imported with
  those links live — clicking one lands on the imported page, not the host page
  sharing its id.

## Size

166,482 → 175,690 B (+9,208, 5.5%). `scripts/size-budgets.json` moves
176,128 → 184,320 in the same commit: the old ceiling left 438 bytes, inside the
packer's own cross-node-version spread (~1.1 KB measured elsewhere), which would
fail CI for no reason.

Touches `docs/DECISIONS.md`, `spaces/CHANGELOG.md` and the i18n catalogs, so
conflicts with the sibling spaces branches are expected there.


nyblnet:

PR #333, opened 2026-08-22
spaces: a table is content, and its cells are ordinary rich text

nyblnet:
## What changed

A `table` block for bento/spaces — a **content** table, per `working/spaces-design.md` §2.6: no formulas, nothing that recalculates, no binding to another document. The database case already shipped here as the tracker (`doc.fields` + `prop` + `view`) and is untouched.

```jsonc
{ "type": "table",
  "rows": [["Name", "Qty"], ["Rice", "2"]],   // row-major, INLINE HTML
  "cols": [2, 1],                             // fractional column weights
  "colAlign": ["", "right"],                  // per COLUMN, not per cell
  "header": false }                           // absent = TRUE
```

- **Model** (`spaces/src/model.ts`) — `tableOf` (one read-time normaliser, never repairs the file), `writeTable` (the one writer, omits every default), `tableFallbackHtml`.
- **Registry** (`spaces/src/blocks.ts`) — one entry: icon, label, slash item, `init`, `toMd`. `MdCtx` gains `inline` so the exporter's html→markdown converter reaches the cells (blocks.ts is DOM-free by rule).
- **Renderer** (`spaces/src/render.ts`) — real `<table>/<thead>/<th scope>`, `<colgroup>` over `table-layout: fixed`, `dir="auto"` per cell. One renderer, so the editor, reading view, print and the still preview cannot disagree.
- **Editing** (`spaces/src/editor.ts`) — contentEditable cells; Tab/⇧Tab and ⏎ walk the grid, Tab off the last cell appends a row, ⇧⏎ is a line break; a bar on the block adds/removes rows and columns and toggles the header; column widths drag.
- **Markdown** — exports as a GFM pipe table with alignment; `spaces/src/markdown.ts` now PARSES one, replacing the code-block workaround whose own comment said it was "mechanically upgradable the day a table block ships".

## The decisions worth reviewing

**A cell is a bare string, not slides' `{html, color, bg, bold}`.** Slides' cells carry presentation because a canvas has no cascade. A space has a theme and a stylesheet, and a spaces cell is already rich text — bold in a cell is `<b>` through the same allowlist as every other block. So the whole `style` object and every per-cell override go away. Alignment stays per column because that is exactly what a GFM rule row can say.

**`header` absent means TRUE** — the opposite of every other boolean in the model. A pipe table always has a header, so that is the case that should need no field, and `{type:'table', rows:[[…]]}` is then a first-class hand- or agent-written document.

**Format additivity is the `html` fallback.** An older build has no `table` case and renders `html`; with none it renders nothing. So `html` is derived from the cells' own inline html (not their text — a `#p/` link in a cell still produces a backlink, still finds ⌘F, still exports as a link) and rewritten by `writeTable` on every edit. Costs a second copy of the table's text; a table that vanishes in the build someone already has is not the cheaper option.

**Under collab, `rows` is one LWW register** — concurrent edits to different cells keep one of the two. Slides' table has the identical documented limitation; the fix is a node per cell, which is a format change, so it is written down rather than half-built.

- **Format additivity (§3)** — new block type, all fields optional, unknown/ragged shapes normalised at read time only, `html` fallback for older builds. Pinned by 40 new assertions.
- **i18n (§8)** — 7 new strings in all eight catalogs, `build-spaces-i18n.mjs --check` complete at 276 × 8. Terminology per catalog (zh-Hans 行/列 vs zh-Hant 列/欄).
- **Splice contract (§2)** — `shell-gate.mjs` passes.
- No kernel change.

## The bug the browser found

The column grips first went inside the first row's cells. A cell's `innerHTML` is the model, so every grip was written into the document as cell content — and `caretToEnd` put the caret INSIDE the trailing `<button>`, so the first word typed into a column landed in the button and was eaten by the sanitizer on blur. Measured: `rows[0][1]` read `<button class="sp-tb-grip">Qty</button>`, then `""`. Grips live in the wrapper now, absolutely positioned over each boundary.

Class names were checked first: `.sp-col` is already the board's column and `sp-col-resizing` is already a body class, so everything here is `sp-tb-*`.

## How it was verified

`npx tsc -b` clean. Rigs: model **504/504** (40 new) · agent 143/143 · calc 90/90 · journal 45/45 · undo 18/18 (via the esbuild bundle CI uses) · i18n `--check` complete · `test-spaces-size.mjs` · `shell-gate.mjs`.

**Real browser**, built shell served under a unique filename, `location.pathname` asserted alongside every measurement:

- inserted a table from the Insert menu — 3 `<th scope="col">`, 6 `<td>`, caret in cell 0,0
- typed across the header with Tab; `rows` and the derived `html` tracked each keystroke
- Tab from the last cell appended a row and landed the caret in it
- removed a row, added a column, removed a column — each from the bar, each one undo step
- toggled the header off (`thead` gone, 9 `<td>`) and back on
- dragged a column boundary: live `<col>` widths, grip following to 137px, committed as `cols [0.583, 1.417, 1]` (total unchanged); ⌘Z restored it
- reading view: 9 inert cells, no editable, no tools, no grips
- print: `.sp-printroot` carries the table with a real `<thead>`, and all six `@media print` rules are live in the shipped stylesheet
- markdown export: `| Name | Qty | Note |` over `| --- | --- | --- |` with the header on; an empty header row with it off

## Size

**170,982 B of 176,128** (97.1%), up **4,500 B** from the recorded 166,482. No ceiling change — that headroom was left for the next feature to spend.


nyblnet:

PR #334, opened 2026-08-22
spaces: one place that says what a block can be, and it stays out of the way

nyblnet:
Based on `spaces-m3m4` (aacf7e2), not `main` — this needs the tables, media and
link cards that only exist on that branch.

## What changed

A right-hand properties panel for bento/spaces (`spaces/src/props.ts`),
structurally the page tree's mirror image: the same `.sp-resizer` strip, the
same always-visible chevron, drag-to-resize, double-click-to-reset, and state
that lives in `localStorage` and never in the file.

**Block properties**, for the block the caret or the last click was in:

| type | what it exposes |
|---|---|
| table | rows, columns (steppers, clamped to `TABLE_MAX_*`), header row |
| code | language, including an unrecognised tag kept and offered rather than silently deleted |
| callout | tone, and the icon that overrides the tone's mark |
| image | width % of the column, alt text |
| media | width and poster (video only — an `<audio>` is a control bar), muted, loop, controls, replace |
| link | url, title, description, site, icon, and the picture dialog |

Anything else says so plainly instead of showing an empty panel, and an unknown
block type spells its own name out.

**Page properties**: icon (hands off to the existing picker), width
(Column/Wide/Full — the page menu's own control, in the place the question is
now asked), archived, the journal date when there is one, and a derived
blocks/words/backlinks line.

Sections are `.sp-insp-sec` headers retrofitted into an accordion after every
rebuild, open state persisted per title — slides' `applyAccordion`, so a section
added later is collapsible without anyone remembering to make it so.

## How it is kept from costing width

This is the part that matters, given the wide-page fix that shipped one commit
earlier.

- **Default CLOSED.** `inspClosed = true`, and only an explicit `'0'` in
  `bento-sp-insp-closed` (written when somebody opens it) opens it. An absent
  key, a new browser, or a `localStorage` exception all mean closed.
- **Closed it takes no width**: `flex-basis: 0`, no inline padding, no border,
  so `.sp-main`'s `flex: 1 1 auto` takes the width straight back.
- **Measured against the same shell without this change**, panel closed:

  | | reading column | `.sp-main` |
  |---|---|---|
  | 1280px, before | 720px | 1031px |
  | 1280px, after | **720px** | 1026px |
  | 2560px, before | 900px | 2311px |
  | 2560px, after | **900px** | 2306px |

  The reading column is untouched at both widths. The whole cost is the 5px
  chevron strip, which is what buys a control on the panel's own edge instead of
  one in a toolbar across the room.
- **Below 820px it is an overlay**, `position: fixed` with the page list's own
  `.sp-scrim` to tap away — not a third column.
- **Its topbar button exists only where the chevron cannot.** Hidden above
  820px (the strip carries the control); shown below; and hidden again once
  `fitTopbar` folds the bar, where it moves into ⋯. Measured at 375px: leaving
  the 40px button in a folded bar took the document title from 70px to **26px**.
  With the demotion the title is 70px, byte-for-byte the pre-change bar.

## Undo

Every control is one `store.commit`; every text field commits on `change`, not
`input`. Verified in the browser: setting an image's width then its alt text
takes exactly two undo entries, each reverting only its own field.

## Verification

- `npx tsc -b` in `spaces/` — clean.
- `node scripts/test-spaces-model.ts` — **653/653** (24 new checks: the
  default-closed preference, the three properties of the closed geometry,
  `.sp-main` being `flex: 1 1 auto`, nothing about the panel reaching the
  document, the overlay below the breakpoint, the accordion's persistence, the
  one-commit rule, and that the block chips survived the panel).
- `node scripts/test-spaces-agent.ts` 155/155 · `…-calc.ts` 90/90 ·
  `…-journal.ts` 45/45.
- `scripts/test-spaces-undo.ts` through the esbuild bundle CI uses — 25/25.
- `node scripts/build-spaces-i18n.mjs --check` — 388 strings × 8 locales, all
  complete. 22 new strings in all eight catalogs; 5 of them lifted verbatim from
  the slides catalogs, which already say "Show/Hide properties (])" because
  slides binds the same key to the same panel.
- `npm run build:single` + `node scripts/shell-gate.mjs` — splice conformance
  OK, runtime ships deflated only.
- **Real browser**, uniquely-named served copy (`props-panel-a6.html`,
  `location.pathname` asserted alongside every measurement):
  - first load with the key cleared: panel closed, 0px wide, chevron reads
    "Show properties (])";
  - opened it, changed a **code block's language** (js → py: model, the language
    chip and the highlighting all follow; one ⌘Z restores it and the panel
    re-syncs);
  - inserted a **table** from the Insert menu — the panel picked it up
    automatically; + on Rows took 3 → 4 in the model and the DOM, Header row off
    left zero `<th>`, one ⌘Z restored the header and left the row alone;
  - an **image**'s width (100 → 50, the on-block chip agrees) and alt text (the
    `` follows), two independent undo entries;
  - a **page**'s width (normal → wide in the model, one undo step) and the
    reading-view lock (all panel controls `disabled`);
  - `]` toggles the panel and is inert while the caret is in a block;
  - accordion state and open/closed state both survive a reload;
  - 2560px and 375px as tabulated above; no horizontal document overflow at
    375px; no console errors.

Not run: `scripts/test-sync-spaces-session.ts`, which needs `slides/node_modules`
for its esbuild resolver hook and is not installed in this worktree. Nothing
under `sync/` is touched.

## Size

196,950 → 203,654 B (+6,704, 3.4%). `scripts/size-budgets.json` raised
204,800 → 212,992 in the same commit with the measurement in its note: 1,146 B
of headroom sits inside the packer's own cross-node-version spread, which the
existing note says fails CI for no reason.

## Platform invariants

None moved. Nothing about the panel is written into the document — it is
viewer state in `localStorage`, the same rule the page-list width, the reader
width, the language and reduced motion already follow. The splice gate passes.
Chrome is laid out with logical properties so it mirrors under RTL.

## Overlap

Deliberately no inline text formatting — a sibling branch owns the floating
selection toolbar for inline marks. Expected conflicts with it in `editor.ts`,
`styles.css`, the eight catalogs, `CHANGELOG` and `DECISIONS`; not worked
around.


nyblnet:

PR #335, opened 2026-08-22
spaces: typing `[` put no character in the block and closed the page list

nyblnet:
`[` collapses the page list, as in slides. The comment beside it said it only reached there when nothing was being edited, because *"the text path returns above"*. **The text path is about ninety lines below it.** So every bare `[` was caught by the shortcut first, `preventDefault()`ed, and turned into a sidebar toggle — the character never arrived.

Which means **`[[` could not be typed at all.** That is how this app makes links, and the starter space tells you to do it in as many words: *"Type `[[` anywhere to search your pages and drop a link in"* — on a page that then demonstrates the result. Shipped since #237.

## Measured, before and after

On a build of `main`, before changing anything:

| | |
|---|---|
| `defaultPrevented` | **true** |
| block text | unchanged |
| `.sp-side` | gained `sp-pane-closed` |

After the fix: the character arrives, the sidebar does not move, and `[` with nothing focused still toggles the panel as it always did.

## The guard

It asks the same question the text path asks, and asks it about **any editable target** rather than only a block host — the page title, a table cell and a comment box all deserve their brackets. A shortcut that eats a character is worse than one that is missing.

## Pinned twice

The specific binding, and a **rule**: every unmodified single-character binding in `onKey` must consult the guard, so the next bare-key shortcut cannot reintroduce this.

That second assertion failed against itself first — `[^)]*` stopped at the parenthesis inside `editingText()`, so it matched nothing and passed vacuously. It reads to end-of-line now.

## Provenance

Found by the agent building the properties panel, which correctly left it alone as not its concern.

Gates: model 469/469 · agent 143/143 · calc 90/90 · journal 45/45 · undo 18/18 · i18n complete · shell-gate OK. No size change worth the name.

### DIFF
--- a/scripts/test-spaces-model.ts
+++ b/scripts/test-spaces-model.ts
@@ -1628,5 +1628,36 @@ for (const [label, input, err] of [
     'Store.tree carries a visited set')
 }
 
+// ---- A BARE-KEY SHORTCUT MUST NOT EAT A CHARACTER ---------------------------
+// `[` collapses the page list. It was unguarded, and the text path that would
+// have claimed the key first sits ~90 lines BELOW it — so every `[` typed in a
+// block was preventDefault()ed into a sidebar toggle. `[[` is how this app
+// makes links and is what the starter space tells you to type; it could not be
+// typed at all. Shipped since #237, found while building the properties panel.
+{
+  const fs = await import('node:fs')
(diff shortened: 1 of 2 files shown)


nyblnet:

PR #336, opened 2026-08-22
spaces: select some words and say how they should look

nyblnet:
Select any words in a space and a small bar appears above them: bold, italic,
underline, strikethrough, inline code, highlight, colour, link, clear
formatting. Underneath it is the Range mark engine that had to exist first.

**Branched from `spaces-m3m4` (aacf7e2), not main.**

## What was wrong

`⌘B`/`⌘I`/`⌘U` existed, nothing announced them, and there was no strikethrough,
inline code, link, highlight or colour at all. The three that existed called
`document.execCommand`, which §2.4(b) forbids — which is how contentEditable's
own markup was reaching the model. And `canonicalize` implemented one third of
§2.3: it coalesced adjacent runs and could not touch nesting order, so
`<b><i>x</i></b>` and `<i><b>x</b></i>` were both canonical — one visible
sentence, two byte spellings, a conflict in every diff and every future merge.

## The mark engine

`spaces/src/marks.ts` is a **pure string function** over (inline html,
plain-text offsets), not DOM surgery — because the rigs are plain `node` with no
DOM, and the case that goes wrong quietly (un-bolding the *middle* of a bold run)
has to be pinned somewhere other than a browser session. The DOM appears only in
`offsetsOf`/`rangeAt` at the bottom of the file. The toolbar holds **offsets**,
never a Range, so nothing stale can survive the `innerHTML` write.

Canonical nesting order, outermost first:

```
a > mark > span > strong > em > u > s > sub > sup > code
```

`b`/`i` fold to `strong`/`em`. The serializer that produces canonical output *is*
the canonicaliser, so the toolbar's output and the commit's normalisation cannot
drift. Verified in the built shell: after four operations the blur
canonicalisation is a byte no-op.

## Colour — the format's second attribute

Notion's shape: nine colours in two roles, `sp-fg-<name>` on a SPAN for the ink
and `sp-bg-<name>` on a MARK for the band. A palette and never a picker, because
a closed vocabulary means the sanitizer matches a *name* instead of parsing CSS.

The allowlist is `/^sp-(fg|bg)-[a-z0-9-]{1,16}$/` — a **pattern**, not the nine
names. With no server to migrate anything, an enumeration would delete a tenth
colour added in two years, silently, on the next edit touching the block. Its
own `DECISIONS.md` entry.

Colour **prints**, deliberately the opposite of the callout ruling in the same
stylesheet: a callout keeps its box, rule, shape and name, so dropping the fill
costs nothing; an inline colour has no second cue.

## Markdown

Every mark exports and imports back. `htmlToMd` was a DOM walk in `about.ts`,
therefore untestable, therefore quietly dropping `u`, `mark`, `sub` and `sup` to
plain text. Moved into `marks.ts` over the same run list.

| mark | markdown |
|---|---|
| `strong` / `em` / `s` / `code` | `**x**` `*x*` `~~x~~` `` `x` `` |
| `mark` | `==x==` — native Obsidian syntax and a Pandoc extension, so it round-trips outside this app too |
| `u`, `sub`, `sup`, colour | raw inline html (GFM permits it; exact round trip) |

`_x_` for underline was rejected: it re-imports as *italic*, and a mark that
comes back as a **different** mark is worse than one that comes back as nothing.
Nine marks × (export, re-import) asserted in the rig.

## Decisions worth arguing with

- **A toggle closes the typing run.** A run takes one checkpoint, at its first
  keystroke; fold a toggle in and the `⌘Z` meant to undo the bold undoes the
  sentence too. `store.commit`, `scope:'page'`, `structure:false`.
- **Touch gets a docked bar.** On `pointer: coarse` the buttons sit at the bottom
  edge, clear of the selection handles and the OS callout. "Nothing at all" was
  rejected — a phone has no `⌘B` either. Measured at 375px: one row, 51px.
- **`⌘K` is two commands**, split on whether anything is selected (Notion's
  split, so no new shortcut). `⌘S` grew a `!shiftKey` test it never had, or
  `⇧⌘S` would open the save dialog.

## Two bugs the browser pass found

Neither reachable by types or by the rigs:

1. `lenOf()` treated "not an element" as zero characters, and it is only ever
   called on `cloneContents()` — a `DOCUMENT_FRAGMENT`. Every selection measured
   `0..0`, the toolbar never appeared, nothing reached the console.
2. Focusing the link field moved the selection out of the block, fired
   `selectionchange`, and closed the bar the field was inside, one frame after
   it opened.

## Verified

In a real browser on a uniquely-named served copy, asserting
`location.pathname` alongside every measurement: bar above the selection and
clamped inside the viewport; all six marks on and off, each landing canonical in
the stored html; a partial un-bold splitting into three runs with the selection
intact and the DOM matching the model; `⌘K` linking, and opening search with
nothing selected; the palette writing `mark > span > strong`; clear formatting;
**nothing in reading view** under a real non-empty selection; the print rules
present.

```
model 697/697 · agent 155/155 · calc 90/90 · journal 45/45 · undo 25/25
i18n 401 × 8 complete · tsc clean · shell-gate OK
206,446 B of 212,992 (96.9%)
```

Ceiling moved from 204,800: 196,950 → 206,446 left 1,210 B of headroom, less
than the 1,151 B the budget file already measures between two node versions of
the same commit.

A sibling agent is building the right-hand properties panel on the same base;
conflicts in `editor.ts`, `styles.css`, the catalogs, CHANGELOG and DECISIONS
are expected and resolved centrally.


nyblnet:

PR #337, opened 2026-08-22
spaces: the About dialog answers the questions slides' does

nyblnet:
It was a flat stack of eleven controls under three style hooks. Slides' has
eighteen, and the difference was not decoration: this one could not say how
big the space was or where the weight went, could not check for updates
without lying about what the button did ("Update this file" downloaded a
copy), showed no release notes, had no auto-check switch, offered no way back
in from JSON, no duplicate, and named the document nowhere. Two apps in one
suite were teaching two different dialogs.

Rebuilt on slides': same opening mark, same section order, same register.
Sections with real headings instead of one long column — This file, Document
properties, Updates, Language, Password, Bring notes in, Take it elsewhere,
and Careful.

WHAT IS NEW. A content readout (pages, blocks, words, document bytes) and the
sentence that explains it — a space is big because of its images, and "4.2 MB"
without that has told you nothing you can act on. Words counts CJK characters
individually, because splitting on whitespace calls a Japanese paragraph one
word. The update section grew the auto-check switch, signed release notes
filtered to the versions THIS file skipped, and a real in-place update with a
rollback copy; the old button said "Update this file" and downloaded one.
Document properties names the space and shows its docId and last save.
Replace-from-JSON came back, and Duplicate as a new space, which mints a fresh
docId and drops the collaboration credentials so a duplicate can never sync
with its parent.

Copy document JSON now strips `collab` before it reaches the clipboard. Every
field under it is a bearer capability — the room, the read key, and the private
halves that grant writing and revoking — and it was copying the live document
whole.

WEIGHT WHERE IT BELONGS. Removing the password and replacing from JSON open an
in-flow panel that states the consequence and puts the verb on the second
button. IN FLOW is the whole point: the card scrolls, and `overflow-y: auto`
clips horizontally too (hard-won #10), so a floating child of it would be cut
on the left and the bottom with nothing to show for it. Measured: both panels
`position: static`, both inside the card's box at 375px, `scrollWidth ==
clientWidth` throughout.

TWO BUGS FOUND WHILE MEASURING. The wordmark rendered as three stacked lines
("bento" / "/" / "spaces") from the day it shipped: `.sp-about-logo span` was
meant for the version line and was also matching the peach slash INSIDE the
`<b>`, and a block in the middle of a line breaks the line on both sides.
`b + span` is the whole fix. And `.sp-btn` is transparent by design — the
topbar is icon-only — so in a white card with no hover every button read as
plain text; the dialog gives non-primary buttons their border back, exactly as
slides' about does.

UPDATES AT LAUNCH. The switch had nothing to switch: spaces only ever checked
by hand. It checks once at launch now, on the kernel's `bento-auto-check` key,
which is slides' key — so turning it off in one app turns it off in both. This
sits beside "A space does not phone home when it is opened", and does not
contradict it: that rule is about the DOCUMENT reaching an author-chosen host,
this is the APP asking its own release origin under the reader's switch, with
no id and nothing from the document. DECISIONS carries the argument.

I18N. 49 new keys across all eight catalogs, 29 of them lifted verbatim from
slides' — which is both cheaper and what keeps the two dialogs sounding like
one product. 467 strings × 8 locales, all complete. Eleven strings the sweep no
longer finds were dropped from the catalogs; they were already absent from
packed.ts, so nothing shipped changes.

VERIFIED in a real browser on a uniquely-named served copy, asserting
location.pathname beside every number. At 1600: card 560px, all eight sections,
no element outside the card's box, fits the viewport. At 375: card 351px with a
12px gutter, no horizontal scroll on body or card, stats reflow 2×2, the whole
dialog scrolls to Close. Focus trapped both directions (16 focusables, Tab off
either end wraps), Escape closes with focus outside the card. Bad JSON refused
without touching the document (11 pages before and after), title field trims and
commits, auto-check persists both ways, Japanese renders every section with
nothing overflowing. Release notes hang their indent: bullet at x=70, wrapped
lines at x=75, which is the flush-left second line slides' releaseNotes() exists
to prevent — and notes are textContent there and here, because a signature says
who wrote a string, not that it is safe to run.

Shell 213,434 → 229,634 B (+16,200), ceiling raised 221,184 → 235,520 with the
measurement in the note. Nearly all of it is language: packed.ts alone grew
22,644 B raw, against roughly 4 KB for the dialog's own code and CSS.

Gates: tsc -b clean (spaces and kernel); model 722/722, agent 155/155, calc
90/90, journal 45/45, offline 33/33, undo 25/25 and sanitize 138/138 (both
bundled with esbuild as ci.yml does); i18n --check complete; build:single and
shell-gate clean; size within budget.


nyblnet:

PR #339, opened 2026-08-22
A dark interface for bento/spaces, page and all

nyblnet:
About → Appearance offers **Match my system**, **Light** or **Dark**, following the
OS by default and tracking it live. The theme is a VIEWER preference
(localStorage, `kernel/src/theme.ts` — the one slides shipped in #285), never
document data. The saved `<html>` tag was measured bare: no `data-theme`, no
`color-scheme`, so nothing about the reader travels with the file.

### Where this diverges from slides, deliberately

slides themes only the chrome *around* the deck, because a slide is a bounded
page on a canvas whose background is the author's data. A space has no such
page: `--bg` is THE ONE GROUND, and the bar, both panels and the reading column
sit on it with hairlines and nothing else between them — which is what the
chrome-reads-as-one-document commit underneath this one settled. A dark window
with a white reading column is a white rectangle over most of the screen, and a
re-introduction of the floating-card look that work removed. **So here the
reading surface IS chrome and follows the reader.**

`doc.theme.background/color/accent` remain the DOCUMENT's colours and remain
untouched. Nothing in the live app has ever painted them; their only consumer is
the static file-manager preview, which keeps them in both themes. That surface
renders with scripting off — no localStorage, no reader, no preference to
honour — so a media query there would be a guess made on someone's behalf.

Read from the other side, the same ruling says the document's colour
*vocabulary* must follow the reader. `sp-fg-*`/`sp-bg-*` and the callout tones
are a closed set of NAMES — the author picked "red", not `#c0392b` — which is
exactly why the value under each name is the stylesheet's to tune per surface.
The five tones collapsed to ONE HUE each, with the fill mixed into `--bg` and
the label into `--ink`: five numbers instead of thirty, adapting both ways.

### Three states, and how paper stays light

Dark is stated under the guarded `@media (prefers-color-scheme: dark)` **and**
under `[data-theme="dark"]`, or the picker only wins in one direction. Verified
with the attribute removed: dark OS → dark, light OS → white,
`data-theme="light"` under a dark OS → white.

Both blocks are `@media screen`, which is how print stays light with **no third
copy of the palette** — it matches neither and falls to the unconditional light
block. The print sheet carries one line the media query cannot,
`color-scheme: light only !important`, because theme.ts writes `color-scheme`
inline on `<html>`.

### The audit

Converted: `#fff` on the bar, focused title, current tree row, seven popovers
and menus, the About/gate cards, five form surfaces, the page and link cards,
the table tools and the pane tab; `#b9c2ce` placeholders and `#aab4c2` gutter
ghosts; five `#b9c3d2` hover borders; `#5d6b80`; `#c0392b`; `#fff6dd` and the
`rgb(247 166 0 / …)` accents; `rgb(255 255 255 / .94)` on three floating tool
rows; `rgb(91 141 239 / .45)` on both resize grips; and every
`rgb(20 30 45 / …)` shadow and scrim → `rgb(var(--shade) / …)`, channels rather
than a colour, so eleven shadows keep their own alphas behind one token that
goes to pure black in dark. Two `color: #fff` on buttons *filled* with `--ink`
became `var(--bg)` — white on near-white the moment the theme flipped. Three
literals stay and are named in the gate: a picture's ground, a clip's letterbox,
and the initial on a generated presence dot.

**An image now sits on a white card of its own in both themes.** Document
content is never inverted — but a diagram exported against a transparent
background is a shape floating in nothing on a dark page.

### Verified

Chrome, both themes, on a uniquely-named served copy with `location.pathname`
asserted beside every measurement. Contrast, light / dark: body text 14.5/14.8,
muted 5.98/5.88, sidebar 9.9/9.3, code roles 5.6–6.9/5.2–8.7, the nine inks
4.7–6.8/6.5–9.1, table cells 13.5/13.1, callout labels 4.98–6.49/6.5–8.4.
Nothing below AA either way; the tone-label mix went 55% → 50% because *warning*
measured 4.52 at 55%.

Gates: `tsc -b` clean; model 738/738 (twelve new theme checks), agent 155/155,
calc 90/90, journal 45/45, undo 25/25, sanitize 138/138, i18n `--check` 429×8
complete, `shell-gate.mjs` OK, `test-slides-theme.mjs` 9/9 and `test-preview.ts`
28/28 unaffected.

**Shell 213,434 → 215,002 B (+1,568, 0.7%). No budget raise** — the ceiling
stays 221,184.

### Merge note

`about.ts` is touched by exactly two lines (one import, one `card.append`); the
control lives in a new `spaces/src/appearance.ts`, to stay out of the in-flight
About rebuild's way. Conflicts in `styles.css` and the i18n catalogs are
expected.


nyblnet:

PR #341, opened 2026-08-23
spaces: an invite handed over the room's owner key — and the rest of the sharing surface

nyblnet:
## The bug

`spaces/src/editor.ts` wired the Live popover's "Invite someone…" to
`invite: () => { void this.saveAs('copy') }`, which reaches `serializeAuto(store.doc)`
in `main.ts` — the whole document, `collab` included.

`collab.ownerPriv` is the room's **root key**: it signs writes and it signs
revocations. So everyone invited to a space received the power to write (intended)
and the power to **revoke the person who invited them, and everyone else** (not
intended, and not recoverable — the file is already on someone else's disk; the only
remedy is re-keying the room and re-sending every copy). The hint under the button
read "Saves a copy that joins this session": true about the mechanism, silent about
the power.

Measured on this branch before the fix: the bytes that button wrote contained
`ownerPriv` in full. Nothing about the copy looks wrong — it opens, it syncs, it
behaves exactly like a correct invite.

**The fix is the first commit**, on its own, so it can be cherry-picked:
[`spaces: an invite handed over the room's owner key`](../../commit/).

## What an invite carries now — and what it does not

`spaces/src/share.ts` is the single place that answers this.

| | invite | view-only | template / clipboard / page extract |
|---|---|---|---|
| `room`, `key` (read cap) | ✅ | ✅ | ❌ block dropped whole |
| `owner` (PUBLIC key) | ✅ | ✅ | ❌ |
| `writerPub` | ✅ | ✅ | ❌ |
| **`ownerPriv`** | **❌** | **❌** | ❌ |
| **`writerPriv`** | **❌** | **❌** | ❌ |
| `invite` (delegation keypair) | ✅ **owner-signed, fresh per copy** | ❌ | ❌ |
| `sync` (stamped CRDT state) | ✅ (copies rejoin as true forks) | ❌ | ❌ |

An invite carries an **owner-signed delegation keypair** — the owner signs
`inv.${pub}.${role}.${exp}`, and every device that opens the copy mints its *own*
member key and joins through the owner → invite → member chain. The owner keeps two
things they did not keep before: one device can be removed without disturbing anyone
else, and revoking an invite cuts off every copy descended from it.

Strip first, then delegate — a stray `writerPriv` from a pre-v2 mint beside an invite
would be a second, **unrevokable** way into the room. The stripper works by
**deleting**, so a private field added to `CollabCreds` later is covered without
anyone acting.

`docForExport()` is untouched and stays as it is. It is the opposite case and the two
must not be merged: the clipboard copy must carry **no** capability; an invite must
carry **exactly one**.

## Which of the six landed

**All six.**

1. **Invite to edit** — owner-signed, revocable, not the owner key.
2. **View-only copy** — keeps the room + read key, holds **no private half at all**.
   With no signing key the socket presents no `?w=`, and the relay stores and fans out
   nothing it sends. The copy also opens locked with a banner (`collab.role: 'reader'`
   → `store.readOnly`) — a *third* reason to lock, distinct from `doc.readonly` (a
   sealed reading copy, no session) and `frozen` (a file this build cannot read). Only
   this one keeps receiving; the lock is a courtesy to the reader, the relay is the
   enforcement.
3. **People list with per-member Remove** — role + key fingerprint (rendered exactly as
   slides renders it, so two people can compare codes over a call), your own row
   included. Remove is an owner-signed `rev.${pub}` that drops **one device**.
4. **Reset access** — quiet at the bottom, and named as the nuclear option: Remove is
   the scalpel, rotation re-mints the room and every copy already sent stops syncing.
5. **Reconnect without saving a copy** — "Stop sharing" had no counterpart, so the only
   route back into a session was to save another copy, which is how one space becomes
   four files.
6. **Honest connection states** — five, not two: offline / view-only / live /
   connecting / off.

## ⚠️ Relay dependency: NONE

**No change to `server/` or `kernel/` is in this PR, and none is required.** v2 rooms,
the owner → invite → member chain, per-socket key pinning and `rev.${pub}` are
**already deployed and already verified by the worker** (`docs/collab-design.md`,
"Phase 1 wire format"; `server/sync-worker/src/worker.js` lines ~336–380 and ~450–470).
bento/spaces was simply not using them — it shares the kernel transport with slides,
which has been minting these rooms since v1.0.3.

The ordering rule (**deploy the relay before a client that depends on it**) is unchanged
and did not bind here. I verified the enforcement path against the **live relay** rather
than assuming it — see below.

## A defect only a browser could find

Measured against the live relay: the Live button read **"Connecting…"** while the panel
two inches under it read **"Live — 1 connected"**. Both call the same `state()`. The
panel is rebuilt every time it opens; the button was painted once at boot, because
`onPeersChanged` was the only caller of `sync()` — and in a room where nobody else has
arrived yet, that never fires. The transport's `onStatus` drives it now, re-armed after
`goLive()` since a transport made just then is a different object with its own callback
slot. Honest states are worth nothing if they are painted once.

## Rigs

**`scripts/test-spaces-invite.ts` (new, 34 checks)** — the assertion this whole task
exists for. Real WebCrypto keys, the real `share.ts` functions, asserting on the bytes:

- the open document really *does* hold `ownerPriv` (the premise — otherwise the rig is
  green for the wrong reason);
- **the owner private key appears NOWHERE in an invite** — the whole serialization is
  scanned, not the field, so a key smuggled under another name or inside a stamped sync
  blob fails too;
- the invite is a *different* keypair, and the owner's signature over it **verifies the
  way the relay verifies it** (a field named `sig` proves nothing on its own);
- each invite is fresh, so revoking one does not cut off another;
- minting one does not disturb the open document;
- a view-only copy holds no private key of any kind and drops its `sync` state;
- private fields are derived from the **kernel** `CollabCreds` type, so a new `*Priv`
  field fails this until it is covered;
- a member copy **cannot mint invites** — otherwise "you stay the owner" would be false.

Bundled (the kernel transport uses TS parameter properties, which node's strip-only
loader refuses), wired into `ci.yml` beside the other bundled spaces rigs.

**`scripts/test-export-secrets.ts` (+15 checks)** — the cheap half, so a *new* button
cannot route around the first rig: a share copy is written from a derived document,
never `store.doc`; it never reaches `saveAs`; the hook uses `serializeAuto` (a password
reaches the shared copy) and does not retain the file handle (or the next ⌘S would
overwrite the copy that just left with the full document). Its body parser could not be
reused as-is: its declaration pattern also matches any two-space-indented `name(`, which
in a module file is a **call** — `stripCollabSecrets(out, …)` inside another function
was being handed back as the stripper's own body, so the checks would have passed
vacuously. `exportedBody()` anchors on `export function`.

## Verified in a real browser

On a **uniquely-named served copy** (`location.pathname` asserted beside every
measurement — the pane caches), **three tabs of one room against the real relay**
(owner / invited editor / view-only):

- ✅ the invited copy joined through the **owner → invite → member chain**, minted its own
  device key in `bento-member-<docId>`, and was not read-only;
- ✅ edits crossed **both** directions (`WRITER-EDIT-9931` → editor and viewer;
  `INVITEE-EDIT-7715` → owner), DOM repainted;
- ✅ People showed **Owner · fingerprint (you)**, **Editor · fingerprint** with ✕, and
  **Viewer** *without* ✕ — correct, a reader carries no key to revoke;
- ✅ **Remove** closed that socket; the removed copy logged
  `[bento-sync] this copy's access was revoked by the owner` and could not reconnect
  (stuck honestly at "Connecting…", never claiming Live again);
- ✅ the view-only copy opened **locked** with its banner, and its write attempt was
  refused (`{ok:false, err:'readonly'}`);
- ✅ **Stop sharing** → button correctly reported *"Also open in another window on this
  computer — not shared online"* (peers locally, not online — the three-state honesty);
  **Start live session** reconnected without saving anything.

**What I could NOT verify this way, stated plainly:** that a revoked or view-only copy's
ops are dropped **by the relay** when the two peers are two tabs of one browser. Same-origin
replicas also sync over `BroadcastChannel`, which is not a relay and is not gated there —
so the revoked tab's later edit still reached the owner locally. That is the **documented
limit**, not a gap in this change (`docs/collab-design.md`, "Threat model / limits":
same machine is one trust domain; read-only is cooperative locally, enforced over the
network). Proving relay-side drop needs two *machines*, or a local `wrangler dev` relay
with BroadcastChannel disabled.

## i18n

31 new strings in **all eight** catalogs, **24 lifted from the bento/slides catalogs word
for word** — every sharing tooltip included. Two apps must not describe one guarantee in
two ways, and rewording costs eight fresh translations to say the same thing. The seven
adapted strings say "space" where slides says "deck".

*Noted in passing, not fixed here (it is slides', not spaces'):* the slides catalogs give
the key `"Editor copy saved — recipients join live with edit access"` the translation of
the **role word** "Editor" in seven of eight locales. The spaces catalogs carry the correct
sentence for that key rather than inheriting the defect.

## Size

| | bytes | |
|---|---|---|
| baseline (`spaces-final`) | 238,862 | |
| after commit 1 (the fix) | 241,250 | inside the existing 240 KiB ceiling |
| after commit 2 | **250,686** | ceiling raised 240 → **256 KiB** |

Measured both ways by building the same tree twice with the old packed catalog swapped
back in: **code +2,308 B**, **language +9,424 B** — four fifths of the raise, which is why
the lifting mattered. Headroom stays deliberate (the packer's output differs across node
versions; a tight ceiling fails in CI for no reason).

## Gates

| gate | result |
|---|---|
| `tsc -b --force` (spaces) | clean |
| `test-spaces-model` | 741/741 |
| `test-spaces-invite` (new) | 34/34 |
| `test-export-secrets` | 45/45 |
| `test-spaces-agent` | 175/175 |
| `test-spaces-calc` | 90/90 |
| `test-spaces-journal` | 45/45 |
| `test-spaces-undo` | 25/25 |
| `test-sanitize` | 138/138 |
| `test-sync-spaces-session` | 48/48 |
| `test-sync-spaces` (250×80×4) | 3/3 |
| `build-spaces-i18n --check` | 504 × 8, all complete |
| `build:single` + `shell-gate` | splice contract OK |
| `test-spaces-size` | 250,686 / 262,144 B (95.6%) |

`scripts/test-spaces-model.ts` gained one assertion and one loosened regex: it pinned the
exact text of `main.ts`'s readonly line, which now has a third reason to lock.

## Notes for review

- `kernel/` and `server/` are untouched. No kernel change was needed.
- `docs/DECISIONS.md` gains one entry (2026-08-23) covering the derive-not-serialize rule,
  the three distinct reasons a space opens locked, the five connection states, and the
  lift-don't-reword i18n rule.
- The five app answers settled 2026-08-18 in `spaces/src/sync/session.ts` are unchanged.
- `spaces-final` did not exist on `origin` — I pushed it at `2caf9bb` (the commit I branched
  from, unmodified) so this PR has a base to target.


nyblnet:

PR #366, opened 2026-08-25
fix(slides): prevent sticky move after fast click

### LINKED ISSUE #260: Selecting an object with single click sometimes causes sticky move
**What happened**
When selecting an object on a slide with a single left click, the object sometimes becomes stuck to the cursor as though clicking and dragging.

**Steps to reproduce**
- Open the Bento Slides Showcase
- Single left-click on objects then move the mouse
- Repro for me is about 1/5 single clicks

**Environment**
- App version (About dialog, or the shell's `APP_VERSION`): v1.0.16
- Browser + version: Vivaldi (Chromium) 8.1.4087.61 (150.0.7871.214)
- OS: Ubuntu 26.04

**Notes**
This only reproduces for me on this older Ubuntu system. Trying to reproduce in the same browser on my M4 Mac fails.

In this video I'm only ever single left clicking to select items.

https://github.com/user-attachments/assets/5f71e6d0-7d24-4ec1-b29f-b472de9be9ce

hamedrabah:
**What & why**

Fixes #260.

A fast single click can release while `waitToChangeTarget()` is still pending. The code can then pass the stale mouse-down event to Moveable after mouse-up. Moveable starts a drag without seeing the matching release, so the selected object follows the cursor.

This change installs a capture-phase mouse-up guard before the async wait. It cancels the stale handoff when release wins the race. A normal held-button drag still uses the existing path.

**How I verified it**

- Reproduced the unpatched failure with a real browser mouse. The object moved from `(850,170)` to `(1014,251)` after mouse-up.
- Ran 30 fast-click trials against the patch. All 30 kept the object at `(850,170)`.
- Verified that a deliberate held-button drag still moved the object to `(960,223)`.
- Ran `node_modules/.bin/tsc -b` successfully.
- Ran `npm run build:single` successfully.
- Ran `node ../scripts/shell-gate.mjs dist-single/Bento_Slides.bento.html` successfully.

**Checklist**

- [x] Ran `node scripts/test-sync.ts` if I touched `slides/src/sync/` (not applicable)
- [x] New UI strings added to every catalog in `slides/src/i18n/` (no new strings)
- [x] Document format changes are additive and backward-compatible (no format change)

### DIFF
--- a/slides/src/editor/canvas.ts
+++ b/slides/src/editor/canvas.ts
@@ -1043,10 +1043,30 @@ export class SlideCanvas {
       this.store.select(this.expandGroups(ids))
       if (e.isDragStartEnd) {
         e.inputEvent.preventDefault()
-        // Selecting a new target and handing the same press to Moveable takes
-        // an async target refresh. A fast click can release during that wait;
-        // starting Moveable afterwards leaves it dragging forever because its
-        // matching mouseup has already happened (#260).
+        // Hand this same press to Moveable, so pressing an unselected element and
+        // dragging moves it without needing a second press. Moveable cannot accept
+        // the press until its target has actually changed, and waitToChangeTarget()
+        // resolves only from componentDidMount/componentDidUpdate — the wait is a
+        // render long and there is NO synchronous path to shorten it.
+        //
+        // A fast click releases inside that gap. dragStart then replays a press
+        // whose mouseup has already been and gone, so Moveable begins a drag that
+        // nothing will ever end: the element follows the cursor, and the click that
+        // finally stops it COMMITS the move to the document (#260). Silent, and the
+        // user was only trying to select something.
+        //
+        // So cancel the handoff when the release wins the race. `mouseup`, not
+        // `pointerup` — Gesto listens for mouse events. Capture phase, so a handler
+        // that stops propagation cannot hide it from us. The guarded window is
+        // complete: selectEnd runs inside the mousedown dispatch, so a release
+        // cannot land before the listener exists. Measured at 40x CPU throttle
+        // (the reporter's symptom is hardware-speed dependent): 22/24 clicks stuck
+        // without this, 0/24 with it, deliberate held drags unaffected.
+        //
+        // Do NOT "simplify" this by making the target swap synchronous — that means
+        // reaching into react-moveable's private _checkChangeTargets(). The real fix
+        // is to stop replaying a stale press and start the drag from a live move
+        // event instead; that is a rework of this gesture, not a tidy-up.
         let released = false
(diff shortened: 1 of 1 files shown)


hamedrabah:
@nyblnet, could you review this focused fix for #260 when convenient? The change is limited to preventing drag activation after a completed click, with browser regression verification documented in the PR.

nyblnet:
Thanks for this — and for the detailed verification notes. I reproduced the failure independently before reviewing the fix, and the diagnosis holds up.

I confirmed the mechanism in the vendored source rather than reasoning about it: `waitToChangeTarget()` (moveable 0.53, `dist/moveable.js:23344`) resolves from `_onChangeTargets`, which fires only once Moveable's targets have actually changed — a render cycle later, not a microtask. So the window between the press and the handoff to `dragStart` is as wide as one frame, which is exactly why #260 reproduces on the reporter's older box and never on their M4.

**Matched-pair measurement.** Two shells built from `main` and from this branch, identical trials, real CDP input (trusted events), CPU throttled 40× to emulate the reporter's hardware — logic untouched, only the machine made slow:

| build | fast clicks sticky | held drag still works |
|---|---|---|
| `main` (unpatched) | **22 / 24** | yes |
| this PR | **0 / 24** | yes |

At 20× throttle the unpatched control came out at 1/10, close to the reporter's "about 1/5" — which is some evidence the throttle is emulating their conditions rather than manufacturing a different bug. Sticky trials were unambiguous, e.g. `sd-tile-a` walking (1074,297) → (1124,327) after mouseup.

One note for anyone repeating this: my first attempt ran in a visible-but-unfocused Chrome window that reported `visibilityState: "hidden"` and **0 rAF frames in 600 ms**. Since this bug is a rAF race, that environment returns "not sticky" for patched and unpatched alike — a pass that means nothing. The headless rig reports `visible` with 30 frames/500 ms, and the held-drag case moves the element exactly as commanded, so the harness is demonstrably capable of moving things.

**Gates.** `tsc -b`, kernel `tsc`, `build:single` and `shell-gate` all pass. Shell cost is +48 bytes.

I rebased onto current main and pushed to your branch (`maintainerCanModify`) so CI tests the real merge result; your authorship is preserved. `canvas.ts` hadn't moved on main since your fork point, so the fix wasn't stale.

On the change itself: using `mouseup` rather than `pointerup` is the right call here — Gesto listens for mouse events, and a pointer-event guard would have missed. `capture: true` gets ahead of Moveable's own handlers, held drags are untouched since mouseup hasn't fired, and `once: true` still cleans up if the promise never resolves. No correctness objection from me.

CI is running now. Final merge call rests with @nyblnet.

PR #367, opened 2026-08-25
Reference equality fix.

tikurahul:
- `e.id === el.id` will be true for all copied slides when using the copy to morph workflow.
- Instead what we should be comparing is `e === el`. (given `String`s are immutable)

**What & why**
- Minor bug fix for code copy slide workflow.

**How I verified it**
- `npm run dev` and tested it locally.
- `npm run build:single` and verified the changes.

**Checklist**
- [ ] Ran `node scripts/test-sync.ts` if I touched `slides/src/sync/`
- [ ] Document format changes are additive and backward-compatible
- [ ] Did not bump the version or cut a release (maintainers sign releases)

### DIFF
--- a/slides/src/code.ts
+++ b/slides/src/code.ts
@@ -78,7 +78,7 @@ export function renderCodeInto(pre: HTMLElement, el: CodeElement, doc: BentoDoc)
   // first slide's tokens — same text, zero travel. The element handed to the
   // renderer is the model object itself, so identity is exact. (The original
   // implementation got this right; a rewrite briefly did not.)
-  const slideIdx = doc.slides.findIndex((s) => (s.elements as unknown[]).includes(el))
+  const slideIdx = doc.slides.findIndex((s) => s.elements.some((e) => e === el))
   if (slideIdx < 0) return false
   const states = diffFor(doc, morphKey)
   const slideStates = states.get(slideIdx)


tikurahul:
This will also be addressed by https://github.com/nyblnet/bento/pull/365. 

Feel free to ignore this in favor of your change.

nyblnet:
You found this independently — thank you. We hit the same bug about eight
hours before your PR arrived, while building the starter deck's two code
slides on the duplicate-a-slide idiom, and it landed in #365. So your patch
no longer applied.

Rather than close yours, I've rebased it onto main, because your version is
the better one. The fix that landed reads
`(s.elements as unknown[]).includes(el)` — that cast exists only to satisfy
`includes`, which wants its argument assignable to the array's element type.
Your `some((e) => e === el)` needs no cast. So this now carries the fix and
deletes the cast, and it stays your commit.

I force-pushed to your branch to do that. Say the word if you'd rather I
hadn't done it that way.

Worth recording: your original #259 code had this right, and my rewrite of the
token source lost it. The tests didn't catch it because every scenario in the
matrix used unique element ids with a shared `morphId` — which is exactly the
arrangement that hides this. The starter deck now exercises the shared-id
path, so it can't come back quietly.

PR #377, opened 2026-08-26
spaces: no hard size ceiling, but the number stays in view

nyblnet:
Your call, implemented. The number behind it: **258,454 B against a 256 KiB ceiling — 98.6%**.

In the last round the ceiling started choosing which fixes shipped. A find-highlight refinement was dropped for size; the `aria-current` fix in #376 landed partly because it happened to cost 32 bytes. A ceiling that decides which accessibility fixes are affordable is costing more than it saves.

## Only the failure goes

The rig gains two modes:

- `enforce: true` — the ceiling exactly as before. **The other apps keep theirs.**
- `enforce: false` — tracked against a `reference` watermark, prints the drift, never fails:

```
track spaces    258454 B  (+312 B since reference, no ceiling)
```

The original entry’s point was never the block — it was that *a budget nobody measures is a wish*, and that a breach must not be found by a reviewer weeks later. That survives. The size is in front of somebody on every run and in every CI log, and drift nobody chose is still visible. `--update` moves the watermark just as it moved the ceiling.

## Verified both paths, not the happy one

| case | result |
| --- | --- |
| tracked shell 58KB past reference | reports `+58454 B`, **exit 0** |
| shell with `enforce: true` over `max` | still **FAILs, exit 1** |
| `--update` on a tracked shell | rewrites `reference`, not `max` |

## Written down

`docs/DECISIONS.md` carries the supersession under the original ceiling entry — what went, what did not, and an explicit *do not quietly put it back*: `enforce: true` restores it for any shell, and this decision is about spaces only. A future round wanting the discipline back is a conversation, not a one-line config edit. The CI step is renamed from "size ceiling" to "size" with the same note, so the log does not claim to enforce something it no longer does.

Gates: full spaces suite green (790/790 model, 175/175 agent, 90/90 calc ×4, 45/45 journal ×5, undo 25/25, invite 34/34), shell-gate clean.

Stacked on nothing — independent of #376, which can land in either order.


nyblnet:
Superseded by #378, which carries the same change from a branch cut off current main and without the workflow-file edit. Closing to keep the queue clear.

PR #402, opened 2026-08-29
Scope an svg's own <style>, which nothing was scoping

nyblnet:
**What & why**

An svg element's markup may carry a `<style>`. Its text was sanitized
(`sanitizeSvgCss` — no `@import`, no `expression()`) but never **scoped**, so its
rules applied document-wide and reached every other svg on the page, including
other slides' copies of the same asset.

That is exactly the hazard `scopeCss` exists for — its own doc comment says so —
and hard-won detail #7 in `CLAUDE.md` is about it. Only `el.css`, the model
field, was ever passed through it. The markup path never was.

**Measured, both directions.** Same probe, same Chrome, only the build differs.
Two svg elements on one slide: one carrying `<style>.p{fill:red}</style>`, the
other only a bare `<rect class="p">`.

| build | style text in the DOM | victim svg fill |
|---|---|---|
| `main` (220f7e7) | `.p{fill:red}` | **`rgb(255, 0, 0)`** — leaked |
| this branch | `[data-el-id="svg-styled"] .p {fill:red}` | `rgb(0, 0, 0)` |

The styled element itself is `rgb(255, 0, 0)` in **both**, so scoping did not
cost an svg its own rules — which was the thing worth checking.

**Shape of the change.** `sanitizeSvg` cannot name an element on its own, so the
scope is passed in and stays optional; an unscoped call behaves exactly as
before.

**The comment was actively wrong.** The note above `SVG_TAGS` said `style` was
safe to allow because `scopeCss` kept its rules from leaking. It did not. A
comment asserting a protection that is not there is worse than none, because it
stops the next reader checking. Corrected, with what was actually true.

**Verification**

- `tsc -b` clean; `npm run build:single`; `shell-gate.mjs` passes
- `test-preview` 28/28, `test-layouts` 9/9, `test-export-secrets` 47/47,
  `test-media-autoplay` 6/6, `test-codediff` pass
- Real-Chrome before/after above

**Not covered by a rig, deliberately.** `sanitizeSvg` needs a DOM, so
`test-sanitize.ts`'s node half cannot reach it, and `scripts/` is the ops zone —
filed for ops rather than folded in here. While running the bar I also found
`scripts/test-clipboard.ts` is dead on `main` for the same extensionless-import
reason as `test-slide-store.ts`, and #399 does not cover it. Also filed.

**Checklist**
- [x] Format changes are additive and backward-compatible — n/a, renderer only


nyblnet:
**Correction, and a merge-order conflict with #399.**

First, I was wrong in the description above. I said `sanitizeSvg` has no rig
coverage because `test-sanitize.ts`'s node half cannot reach it. The **browser
half does cover it**, CI has Chrome, and it caught this change immediately:

```
✗ the svg <style> is kept — scopeCss is what stops it leaking
```

I did not run `test-sanitize.ts` locally — the one rig most relevant to the
change — which is why CI found this and I did not. The failure is real, not a
formatting quibble: the check asserts the sheet still contains
`.dot{fill:url(#bp-ga-am)}`, and a scoped sheet reads
`[data-el-id="…"] .dot {fill:url(#bp-ga-am)}`.

**The conflict.** #399 (ops, awaiting merge) rewrites that exact assertion to
state that the markup sheet is deliberately **not** scoped:

> It is NOT scoped: scopeCss has exactly one call site in render.ts and that
> call site is el.css, not the markup.

That is an accurate description of `main` today, and this PR changes the thing
it describes. So the two are in direct conflict whichever order they land:

- **#399 first** — this PR still fails that check, and fixing it means editing
  `scripts/`, which is the ops zone, not mine.
- **#402 first** — #399's new comment and check name become false on arrival.

**This is not mine to resolve**, on two counts: cross-zone conflicts and merge
order belong to the integrator, and merging is maintainer-only. What I would
argue for, as the slides owner: the leak is real and measured, and hard-won
detail #7 exists to prevent exactly it — so the behaviour should change and the
rig should then describe the new truth, rather than the rig freezing a gap in
place. But that is a recommendation, not a decision.

Marked blocked on the board with this dependency named.

PR #403, opened 2026-08-29
Record that a green local rig run can mean half the rig never ran

nyblnet:
Docs only — one entry appended to `docs/DECISIONS.md`.

**The finding.** `scripts/test-sanitize.ts` has a node half (pure functions,
runs anywhere) and a browser half (drives real Chrome, because what it tests
needs a DOM). With no Chrome, the browser half says so, skips, and **the run
still ends green**.

`sanitizeSvg` needs a DOM, so it is covered *only* by the browser half. That is
how a change to it passed five rigs locally and was caught by CI a minute later
with `✗ the svg <style> is kept` (#402).

**The rule:** not "run the rigs" but *run the rig that covers the file you
edited, and check which half of it ran*. Not how many passed — which ones.

It belongs with the instrument-lies family this log already collects: a CSS
transform on an inline element that reads back verbatim and moves zero pixels,
`document.fonts.check()` returning true for an uninstalled font, rAF throttled
to zero in a hidden tab, an SVG rasteriser ignoring `font-family`.

**The correction is the point.** I first filed this as "`sanitizeSvg` has no rig
coverage" — false. The coverage exists and works. "Your green local run may have
skipped it" is true and actionable; "there is no coverage" would have sent
someone to write a rig that already exists.

Raised by `bento-team-lead`, who suggested promoting it rather than leaving it
in the zone notes.

**Note on scope:** `docs/` is scribe's zone, but the session ritual asks every
zone to promote durable findings to `DECISIONS.md`, and this is append-only —
no existing entry touched. Deliberately NOT touching root `CHANGELOG.md`, which
is scribe's and has #400 open on it.

**Checklist**
- [x] Docs only — no code, no format change, no version bump
- [x] Append-only; conflicts with other in-flight `DECISIONS.md` entries are a rebase, not a merge decision


nyblnet:
**Conflict diagnosed, and it is a genuine keep-both.** Re-checked against `main` at `ad18605`.

One conflict, in `docs/DECISIONS.md`, and it is the append-only pattern: both sides added a NEW entry at the same tail position.

- `main` gained **"2026-08-29 — the locale list leaves the prose; docs point at LOCASES"** (via #394's fallout).
- This branch adds **"2026-08-29 — a green local rig run can mean half the rig never ran"**.

They are different entries about different things. Neither supersedes the other, nothing was edited in place, and no text is contested. **Resolution: keep both, `main`'s entry first, this branch's after it.** Same date, so the order is cosmetic; putting `main` first keeps the diff to a pure append.

Worth stating explicitly because the opposite case exists and looks identical from the conflict markers alone: earlier in this batch a `DECISIONS.md` conflict was called "both-sides-appended, keep-both" when `main` had in fact contributed nothing at that hunk, and keep-mine was correct. Check which side actually added text before applying this. Here, both did.

**Not resolving it myself:** `docs/` is scribe's zone, and this PR's own body already flags that. Handing over rather than reaching in — scribe was offline when this was checked. Queued on the board.

For the record, the finding is slides' (this session's), promoted at bento-team-lead's suggestion.

PR #405, opened 2026-08-29
home/ios: the index rig stopped checking the shared corpus at the rename

nyblnet:
`scripts/test-tray-index.mjs` looked for the corpus at `tray/fixtures`. The tray→home rename moved it to `home/fixtures`, and the rig went on printing **"shared corpus: absent, skipped"** — green every run, with the cross-language check silently not happening.

Nothing had drifted, which is luck rather than evidence. Pointed at the real path, the Swift port agrees with all **11 cases** and the budgets.

## The skip was mine, and it was half right

I wrote that skip and argued for it: *a rig that fails until an unrelated branch lands is a rig people learn to ignore.* That still holds for a corpus which does not exist yet. It does not hold for one that **moved**, and the skip could not tell the two apart.

It can now. Before skipping, the rig looks for an `expected.json` in a fixtures directory elsewhere in the tree. Finding one means this path is stale, and that is a hard failure naming where the corpus actually is.

| corpus state | before | after |
|---|---|---|
| in place | ran | runs — 11 cases agree |
| **moved** | **"absent, skipped", exit 0** | **exit 1, names the real location** |
| genuinely absent | skipped | skips, loudly |

All three verified by moving the directory and running it.

## Why this matters beyond one path

The zone brief already says *a rig that only fails on a total revert is not testing the fix.* This is that rule one turn further out: **a rig that skips when its input moves is not testing anything at all, while still printing a pass.** The failure mode is quieter than a wrong assertion, because there is no assertion left to be wrong.

## Incidental, and a repeat offender

The new helper's doc comment originally contained a literal `*` `/` inside a glob, which closed the block comment early and broke the file — the same trap `Accept: */*` set in `Releases.swift` a few days ago. Rewritten to describe the pattern rather than spell it.

## Adjacent, not mine

`scripts/test-doc-index.mjs` still points at `tray/doc-index.mjs` and dies on it — same rename, same cause, ops zone. PR #399 repairs and registers it, and should land first: it is the JS reference this rig's corpus half is diffed against.

## Exercised

The three home rigs (`test-tray-bridge`, `test-tray-index`, `test-tray-releases`), plus a clean `xcodegen` + simulator build of the host. **No Swift changed, so no save round trip was run.**

### DIFF
--- a/scripts/test-tray-index.mjs
+++ b/scripts/test-tray-index.mjs
@@ -312,7 +312,7 @@ function runSwift(bin, files) {
         // `library.js`'s describe() ALWAYS returns a title, falling back to the
         // file's base name, because its caller has already sniffed and it never
         // faces a non-document. The Swift port answers all the questions in one
-        // call (the shared contract in tray/fixtures) so it returns nil, and the
+        // call (the shared contract in home/fixtures) so it returns nil, and the
         // listing supplies the file name. Same behaviour, different seam — the
         // fallback is applied here so the two are compared like for like.
         title: row.title ?? basename(row.path)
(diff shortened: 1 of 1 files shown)


nyblnet:
**Superseded by #399 — do not merge this. It will be closed when #399 lands, not before.**

#399 and this PR fixed the same bug independently and conflict on `scripts/test-tray-index.mjs`, which is ops' zone (`START-HERE.md:51`). My process error: I claimed `home-ios` and edited a file in `scripts/`, so the board could not show two of us in one function and the integrator's pairwise check found it instead. Both PRs read CLEAN throughout, because that field compares each PR to `main` and never to another PR.

**The composed policy lands in #399**, agreed with `bento-team-ops`:

- path → `home/fixtures`
- absent → **hard fail**, no skip branch
- my `strayCorpus()` survives as a **message enricher only** — no control flow, no throw
- `process.exitCode` with fall-through, so the Swift-diff summary still prints before the failure

**Two things I got wrong here, both worth recording:**

**The skip branch was unreachable, not merely wrong policy.** The rig imports `home/webext/src/library.js` (line 203) and compiles `home/ios/BentoIndex.swift` (line 291) before the corpus section at 407+. Remove `home/` and it dies at exit 2 long before it could skip. ops added the precision that a sparse checkout *could* keep `home/ios` and `home/webext` while excluding only `home/fixtures` — so "unreachable" is true of the realistic case, not literally every case. That residual case argues for hard-fail too: a rig silently skipping its own subject because someone excluded the subject is the failure being removed.

**My `strayCorpus()` threw, and the throw aborted the rig's other half.** Measured: with the corpus moved, `swift port agrees` and `documents compared` both print **zero** times, and the reader's last line is a raw Node stack trace. The 72-document Swift diff — this rig's other, independent guarantee — never reports. That is exactly the defect I had described hours earlier in another file: the error withholds the useful work *and* steers the reader toward "this rig is broken" rather than "this path is stale". ops caught it in my code.

**Why this stays open rather than closing now:** #399 is blocked on its maintainer. Closing this first would leave `main` with a rig that silently skips its corpus and no open PR fixing it. This is the fallback until #399 lands.

nyblnet:
**Superseded by #399 — and merging this now would regress `main`.**

The corpus-path repair landed with #399, and the `strayCorpus()` idea from this PR came with it. `main` credits it in the source:

> `From bento-team-home-ios (PR #405); `tray` dropped from the probe roots, the rename having landed.`

`main`'s version then improves on this branch in two ways, both recorded in the comment there:

1. **This PR throws; `main` sets `process.exitCode = 1`.** From `main`: *"Its ancestor in PR #405 threw instead, which aborted the run before the Swift-diff summary printed and left a raw stack trace as the last thing on screen — the rig read as crashed rather than stale-pathed. Diagnosis belongs in the message, not in the control flow."*
2. **This PR still skips when the corpus is absent** — louder (`ABSENT — skipped`), but still a skip. `main` makes absence an outright failure: *"This corpus has landed; absent means a moved or broken path, not 'not yet'."*

So merging this would restore a skip that prints green, which is the exact defect the PR was written to kill. The finding was right, the mechanism survived, and the version on `main` is the better one.

Verified against `main` at `ad18605`: this branch conflicts, and every conflict hunk is `main` carrying a superset of what this side adds.

Closing as superseded rather than merging. The idea and its attribution are in the tree. @nyblnet — reopen if home-ios sees something here that `main` did not absorb; this was closed while that session was offline, on the maintainer's instruction.

PR #408, opened 2026-08-31
platform: inline Lucide external-link icon in preview panel

rynnwang:
Replaces the Unicode `↗` glyph in the preview panel's "Open in new tab" link with an inline SVG — the Lucide `square-arrow-out-up-right` icon, the same icon this is modelled on in Claude's own web UI.

- Added `openExternal` to the `ICONS` map in `demo.ts`, following the existing hand-inlined Lucide conventions (`fill="none" stroke="currentColor" stroke-width="2"`, round caps, 13px to match the link text).
- Two CSS rules make `#previewOpenTab` an `inline-flex` row so the icon aligns with the label.
- README mention updated.

No new dependency — the platform icon set is already Lucide paths inlined by hand.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rynnwang:
Wrong target repo — recreating against rynnwang/bento.

PR #423, opened 2026-09-08
kernel: read the release public key and sync host from AppConfig

johanhal:
## kernel: read the release public key and sync host from AppConfig

`kernel/src/update.ts` hardcodes `PUBLIC_KEY_JWK` and `kernel/src/sync/online.ts` hardcodes `DEFAULT_SYNC_HOST`. Both are per-publisher values, not per-kernel ones: a downstream build that runs its own signed release channel and its own relay has to patch the kernel to change them, and then carries that patch across every merge.

This adds two **optional** fields to `AppConfig`:

- `publicKeyJwk` — the P-256 public JWK manifests and pack indexes are verified against. Absent, `update.ts` uses the embedded platform key exactly as today.
- `syncHost` — the default relay host. Absent, `sync/online.ts` uses `DEFAULT_SYNC_HOST` exactly as today. The `bento-sync-url` localStorage dev override still wins over both.

Nothing changes for the apps in this repo: none of them set the fields, and every existing rig passes unchanged. `test-release-channel.mjs` still swaps the embedded constant by regex, because the constant is still there as the default.

Why optional and why in `AppConfig`: `app.ts` already says only three values are app-specific across the kernel. These two are publisher-specific, which is the same axis, and `configureApp()` is the one call every app already makes before any kernel module runs. `syncHost()` reads the config lazily and tolerates an unconfigured kernel, so rigs that never call `configureApp()` are unaffected.

Verified: `tsc -b`, `tsc -p ../kernel`, `npm run build:single`, `shell-gate.mjs`, `test-release-channel.mjs --app slides`, `test-sync-session.ts`, `test-relay-protocol.ts`, `test-offline.ts`, `test-storage.ts`, `test-autosave.ts`.

### DIFF
--- a/kernel/src/app.ts
+++ b/kernel/src/app.ts
@@ -6,7 +6,9 @@
 // Only three values are app-specific across the whole kernel: the id the
 // release manifest is signed with, the display name that appears in window
 // titles and save dialogs, and where updates are fetched from. Everything
-// else in the kernel is genuinely app-agnostic.
+// else in the kernel is genuinely app-agnostic. Two OPTIONAL values exist for
+// a fork that runs its own release channel and relay: the signing public key
+// and the default sync host. Absent, the platform defaults apply unchanged.
 //
 // Deliberately NOT a general "kernel init" — modules that need no config
 // (anim, charts) take none, so there is no import-order trap and no false
@@ -24,6 +26,16 @@ export interface AppConfig {
   appName: string
   /** Release manifest URL. Dev override: localStorage 'bento-update-url'. */
   manifestUrl: string
+  /** Release-signing PUBLIC key (P-256 JWK) that manifests and pack indexes
+   *  are verified against. Optional: absent, the kernel uses the platform key
+   *  embedded in update.ts. A fork that publishes its own signed channel sets
+   *  this so its shipped files never accept a manifest signed by another key
+   *  (appId already refuses another app; this refuses another PUBLISHER). */
+  publicKeyJwk?: { kty: 'EC'; crv: 'P-256'; x: string; y: string }
+  /** Default relay host for collaboration (wss://…). Optional: absent, the
+   *  kernel uses DEFAULT_SYNC_HOST in sync/online.ts. The localStorage
+   *  'bento-sync-url' dev override wins over both. */
+  syncHost?: string
 }
 
 let config: AppConfig | null = null
--- a/kernel/src/sync/online.ts
+++ b/kernel/src/sync/online.ts
@@ -17,6 +17,7 @@ import { offlineEnabled } from '../update.ts'
 // Every request in the app goes through the one chokepoint (kernel/src/net.ts)
 // so the offline switch cannot be forgotten — see GHSA-5c3x-xqp6-g94r.
 import { netWebSocket } from '../net.ts'
+import { appConfig } from '../app.ts'
 
 /** the app's store, structurally — see session.ts HostStore */
 type Store = HostStore
@@ -144,12 +145,16 @@ export async function mintCollab(): Promise<CollabCreds> {
   }
 }
 
-/** dev override for the relay host (e.g. ws://localhost:8787) */
+/** The relay host: the localStorage dev override (e.g. ws://localhost:8787),
+ *  else the host the app configured (a fork running its own relay), else the
+ *  platform default. Rigs that never call configureApp() get the default. */
 export function syncHost(): string {
+  let configured: string | undefined
+  try { configured = appConfig().syncHost } catch { /* not configured: platform default */ }
   try {
-    return lsGet('bento-sync-url') || DEFAULT_SYNC_HOST
+    return lsGet('bento-sync-url') || configured || DEFAULT_SYNC_HOST
   } catch {
-    return DEFAULT_SYNC_HOST
+    return configured || DEFAULT_SYNC_HOST
   }
 }
 
--- a/kernel/src/update.ts
+++ b/kernel/src/update.ts
@@ -132,7 +132,9 @@ export async function verifySigned(raw: string, what = 'signed file'): Promise<u
     throw new Error(`the ${what} is malformed`)
 
   const key = await crypto.subtle.importKey(
-    'jwk', PUBLIC_KEY_JWK as JsonWebKey,
+    // A fork with its own release channel supplies its key via configureApp();
+    // the platform key below is the default so upstream builds are unchanged.
+    'jwk', (appConfig().publicKeyJwk ?? PUBLIC_KEY_JWK) as JsonWebKey,
     { name: 'ECDSA', namedCurve: 'P-256' }, false, ['verify'],
   )
   const ok = await crypto.subtle.verify(


nyblnet:
Read from the slides side — this is kernel zone, so the merge decision sits with the kernel owner and security, not me. What I checked, and what I'd ask for:

**The trust root stays build-time.** `configureApp()` is called once per app from `main.ts` with a literal; `config` is module-private and not on `window`. Nothing in a document or a URL can reach it. Absent the new fields, both paths fall through to exactly what runs today. The layering reads right: `appId` refuses another *app*, `publicKeyJwk` refuses another *publisher*.

**One asymmetry to fix.** `appConfig()` throws if never configured. You guard that on the sync path — your own comment says *"Rigs that never call configureApp() get the default"* — but not on the update path, where `appConfig().publicKeyJwk ?? PUBLIC_KEY_JWK` would throw before reaching the fallback. It's not a live bug: every current caller of `verifySigned` configures first (I checked test-packs, test-release-channel, and the webext rig, which uses its own copy). But it's the next node tool's foot-gun, and you clearly intended the guard. Same `try { } catch { }` on the update side and it's consistent.

Nothing else from me. Your README's line — *"any further kernel change stops the work and becomes an upstream PR first"* — is the right posture and it's why this one is easy to read.

nyblnet (commented):
Kernel-zone review — mechanism and invariants only. Whether the door should exist is being reviewed separately; this is whether it is built right. It is, with one property to record and one ask on this PR I'd resist.

## What holds

**PLATFORM §6 holds in substance.** The key is still chosen at build time — `configureApp()` is called once from `main.ts` with a literal, compiled into the shell, module-private, unreachable from a document or URL. So an already-shipped file accepts exactly what it was built with, before and after this change. The format is untouched (`#bento-doc` never sees any of this), so additivity isn't in play, correctly.

**§6's wording is now slightly stale.** It says verification is "against the `PUBLIC_KEY_JWK` embedded in the shell"; after this it's "a build-time key, `PUBLIC_KEY_JWK` by default". Same invariant, names the constant. That's a `PLATFORM.md` edit and maintainer-only, so not this PR's — noted for the lead.

**`syncHost` does not touch the room-commitment scheme.** `w` + sha256(writerPub) is computed client-side and is host-independent; the relay verifies the commitment at connect regardless of which relay it is. No interaction.

## One property this PR makes reachable — record it

`CollabCreds` carries `room` and `key` and **no host**. Before this PR every build had one relay, so §5's "the saved file is the capability" was complete. After it, a file shared across publishers opens on the *other* relay: the room id is a valid commitment, so that relay simply creates the room, and both parties are live — on different relays, never meeting. Silent split-brain. Not a security issue (blind relay, E2EE), a correctness one, and the one thing this diff makes possible that wasn't.

I'd **accept and document it** rather than fix it here: a fork is its own trust domain, and this is consistent with the update channel refusing cross-publisher manifests. One sentence on `syncHost()` saying a file's collab capability is complete only within the publisher that minted it, and a `DECISIONS.md` line. The real fix — stamping the host into `collab` — is an additive format change and a separate serialized kernel PR, if cross-fork collab is ever wanted.

## The parity ask — please don't

Slides asked for try/catch on the update path to match the sync path's. **The asymmetry is correctly oriented and should stay.** Signature verification fails **closed** (unconfigured → refuse); relay selection fails **open** to a blind relay (harmless). A try/catch in `verifySigned` falling back to `PUBLIC_KEY_JWK` would make a misconfigured fork build silently accept *upstream* manifests — the exact outcome `publicKeyJwk` exists to prevent.

Also worth knowing: the update path **already threw when unconfigured before this PR** — `verifyManifest` reads `appConfig().appId`. This change just moves the throw one call earlier, into `verifySigned`, whose only other in-tree caller (`slides/src/packs.ts`) runs configured. So the node-tool foot-gun is pre-existing and loud, not new and silent. If anything, a clearer message from `appConfig()` naming what needed configuring.

## Nit

The PR body says `test-release-channel.mjs` "still swaps the embedded constant by regex, because the constant is still there as the default." True for upstream, which sets no field. For a fork that *does* set `publicKeyJwk`, the regex swaps the unused default and the rig verifies against the config key instead — the fork's problem, but the sentence overclaims.

---
Verified against `kernel/src/app.ts`, `update.ts`, `sync/online.ts`, `sync/crdt.ts:177`, `docs/PLATFORM.md` §5/§6/§9. Provenance is the maintainer's; I'm not marking this either way.

nyblnet:
Retracting my one ask above — kernel's review is right and I was wrong.

I asked for the update path to guard `appConfig()` with try/catch to match the sync path. That would be the wrong fix: signature verification should fail **closed**. A fork that set `publicKeyJwk` but somehow ran unconfigured would, with my change, fall back to the *upstream* key and silently accept upstream manifests — the exact outcome `publicKeyJwk` exists to prevent. Relay selection failing open to a blind relay is harmless; verification failing open is not. The asymmetry is the design, not an oversight.

Kernel also pointed out the throw was already there before this PR (`verifyManifest` reads `appConfig().appId`), so this only moves it one call earlier. Pre-existing and loud, not new and silent.

So: nothing further from slides on this PR. Sorry for the noise — the review that caught it is the one worth reading.

nyblnet:
Kernel and security have both reviewed (the trail is in the review above and in the team's handoffs). Verdicts: **built right** (kernel) and **the shape is acceptable** (security). Neither is a blocker. What's left is three conditions, and I've done the two that are code — as one cherry-pickable commit, since `betamobility/slides` is an org fork and I can't push to your branch:

```
git fetch https://github.com/nyblnet/bento.git pr423-publisher-rig
git cherry-pick 61df394
```

**1. The publisher property is now tested** (security's condition). Every existing case configured the app *without* `publicKeyJwk`, so "refuses another publisher" was asserted by nobody — a refactor that quietly stopped reading the override would have made every fork accept upstream manifests with the rig still green. `test-release-channel.mjs` gains three cases: an upstream build refuses a fork-signed manifest; a fork build accepts its own publisher (positive control); a fork build refuses the platform publisher. The "fork" is a second throwaway key signing the *same* payload, so the only difference is who signed. I negative-controlled it by injecting the regression — verifying against `PUBLIC_KEY_JWK` unconditionally — and both fork cases fail (23/25). 25/25 with the override intact.

**2. No in-repo app sets either field** (security's second condition) — a fourth case reads each app's `configureApp()` block from source. Setting one upstream now fails a rig, which makes it a deliberate act.

**3. The split-brain property is recorded on `syncHost()`** (kernel's ask). `CollabCreds` carries room and key and no host; with per-publisher relays, the same file opened in the other publisher's build joins the other relay, the room id is a valid commitment, and both sides go live on two relays never meeting. Kernel and security independently reached the same conclusion: accept and document — a fork is its own trust domain, consistent with the update channel — rather than fix here. Stamping the host into `collab` is an additive format change and its own kernel PR if cross-publisher collab is ever wanted.

The third condition — `PLATFORM.md §6` now describing a build-time key rather than "the `PUBLIC_KEY_JWK` embedded in the shell" — is the maintainer's text and not this PR's.

Kernel also overturned the one ask I'd made earlier (retracted above): verification correctly fails **closed** when unconfigured, and a fallback to the platform key would have been the exact hole `publicKeyJwk` closes. Your asymmetry was right.

Patch below if you'd rather apply it by hand.

<details><summary>the patch</summary>

```diff
diff --git a/kernel/src/sync/online.ts b/kernel/src/sync/online.ts
index 0c157a1..4cca663 100644
--- a/kernel/src/sync/online.ts
+++ b/kernel/src/sync/online.ts
@@ -147,7 +147,19 @@ export async function mintCollab(): Promise<CollabCreds> {
 
 /** The relay host: the localStorage dev override (e.g. ws://localhost:8787),
  *  else the host the app configured (a fork running its own relay), else the
- *  platform default. Rigs that never call configureApp() get the default. */
+ *  platform default. Rigs that never call configureApp() get the default.
+ *
+ *  A file's collab capability is complete only WITHIN the publisher that
+ *  minted it. `CollabCreds` carries the room and the key and no host, which
+ *  was total while every build had one relay; with per-publisher relays, the
+ *  same file opened in another publisher's build joins that publisher's
+ *  relay, where the room id is a valid commitment and is simply created
+ *  fresh. Both sides are live, on two relays, and never meet — a silent
+ *  split, not a leak (the relay is blind and the payloads are E2EE). This is
+ *  accepted and consistent with the update channel refusing another
+ *  publisher's manifests: a fork is its own trust domain. Cross-publisher
+ *  collaboration would need the host stamped into `collab` — an additive
+ *  format change, and its own serialized kernel PR. */
 export function syncHost(): string {
   let configured: string | undefined
   try { configured = appConfig().syncHost } catch { /* not configured: platform default */ }
diff --git a/scripts/test-release-channel.mjs b/scripts/test-release-channel.mjs
index 0caf44e..9f1041d 100644
--- a/scripts/test-release-channel.mjs
+++ b/scripts/test-release-channel.mjs
@@ -285,5 +285,73 @@ await throws(
   'the untouched shell gets PAST the integrity check (and then needs a browser)',
 )
 
+// ---- 4. the publisher is a build-time choice, and it is honoured -------------
+//
+// `AppConfig.publicKeyJwk` lets a fork that runs its own signed channel verify
+// against ITS key instead of the platform key embedded in update.ts. The
+// property that makes that safe — "refuses another PUBLISHER" — was untested:
+// every case above configures the app WITHOUT the field, so a refactor that
+// quietly stopped reading it would make every fork accept upstream signatures,
+// and nothing here would move. Three cases, because either direction alone can
+// be passed by a verifier that refuses everything, or one that accepts
+// everything.
+//
+// The "fork" is a second throwaway keypair. Its manifest is the SAME payload,
+// re-signed — so the only thing that differs between accepted and refused is
+// which publisher signed it.
+console.log('\n--- publisher: the configured key is the one that counts ---')
+const { releaseSigner } = await import('./lib/release-sign.ts')
+const fork = await releaseSigner()
+const forkManifest = await fork.envelope(payload)
+
+// direction 1: an UPSTREAM build (no override) refuses a fork-signed manifest.
+serve({ [MANIFEST_URL]: forkManifest })
+{
+  const r = await kernelUpdate.checkForUpdates(MANIFEST_URL)
+  ok(r.status === 'error' && /signature is INVALID/i.test(r.message),
+    `an upstream build REFUSES a manifest signed by another publisher (${r.status})`)
+}
+
+// now build "as the fork": same app, its own publisher.
+kernelApp.configureApp({
+  appId: app.appId,
+  appName: app.label,
+  manifestUrl: MANIFEST_URL,
+  publicKeyJwk: fork.jwk,
+})
+
+// positive control: the fork build ACCEPTS its own publisher. Without this the
+// case below would pass for a verifier that refuses everything.
+{
+  const r = await kernelUpdate.checkForUpdates(MANIFEST_URL)
+  ok(r.status === 'update' && r.release.version === payload.version,
+    `a fork build ACCEPTS its own publisher's manifest and offers v${payload.version}`)
+}
+
+// direction 2: the fork build REFUSES the upstream (platform) publisher — the
+// case the field exists for. A refactor that ignored publicKeyJwk fails HERE.
+serve({ [MANIFEST_URL]: manifestRaw })
+{
+  const r = await kernelUpdate.checkForUpdates(MANIFEST_URL)
+  ok(r.status === 'error' && /signature is INVALID/i.test(r.message),
+    `a fork build REFUSES the platform publisher's manifest (${r.status})`)
+}
+
+// And no app IN THIS REPO sets either field. They are for a downstream build;
+// an upstream app setting one is a deliberate act that deserves a review, not
+// something that slips in. Read from source so the assertion is about what
+// ships, not about what this rig configured.
+{
+  const apps = ['slides', 'spaces', 'dash', 'type']
+  const offenders = apps.filter((a) => {
+    const p = join(root, a, 'src/main.ts')
+    if (!existsSync(p)) return false
+    const block = readFileSync(p, 'utf8').match(/configureApp\(\{[\s\S]*?\}\)/)?.[0] ?? ''
+    return /\b(publicKeyJwk|syncHost)\b/.test(block)
+  })
+  ok(offenders.length === 0,
+    `no in-repo app sets publicKeyJwk or syncHost${offenders.length ? ` — ${offenders.join(', ')} does` : ''}`)
+}
+
 console.log(`\n${checks - failures}/${checks} checks passed`)
 if (failures) process.exit(1)
```
</details>

PR #425, opened 2026-09-08
fix(editor): keep the layout picker inside the viewport

johanhal:
The New-slide button at the bottom of the sidebar opened the layout picker upward from itself. Once a deck carries a few custom layouts the picker is taller than the space above the button and its top lands above the viewport: measured top = −7 px at a 600 px-tall window, with the first row of thumbnails clipped under the topbar.

Every anchor now opens beside itself, clamped on-screen. The picker is appended first so the clamp uses its real height, and `.ed-layoutpick` caps at `calc(100vh - 16px)` (it already scrolls) so very short windows scroll inside the picker instead of losing rows.

Measured after the change at 1280×600: top 150, bottom 592, every layout reachable, no internal scroll. At 1280×400: top 8, bottom 392, scrolls internally; choosing a layout after scrolling inserts the slide and closes the picker. The insert-gap and panel anchors keep their placement; only the clamp floor changes from a fixed 460 px to the measured height.

Found in the Beta Mobility fork (six custom layouts in every template).

### DIFF
--- a/slides/src/editor/editor.ts
+++ b/slides/src/editor/editor.ts
@@ -1755,17 +1755,15 @@ export class Editor {
       }
       pick.appendChild(grid)
     }
+    // Open beside the anchor, clamped on-screen. The bottom-of-sidebar button
+    // used to open the picker upward from itself, which pushed a picker with
+    // a handful of custom layouts above the viewport (measured: top = -7px at
+    // a 600px-tall window). The height is read after appending so the clamp
+    // uses the real box; the stylesheet caps it to the viewport and scrolls.
     const r = anchor.getBoundingClientRect()
-    if (anchor.classList.contains('ed-add-slide')) {
-      // bottom-of-sidebar button: open upward from it
-      pick.style.left = `${Math.max(8, r.left)}px`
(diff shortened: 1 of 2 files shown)


nyblnet:
Reproduced your measurements exactly, which is not something I get to say often. Six custom layouts added, picker opened from the New-slide button, real headless Chrome:

| viewport | `main` | this PR |
|---|---|---|
| 1280×600 | top 1293, bottom 1713 — **entirely off-screen** | top 150, bottom 592, no internal scroll |
| 1280×400 | — | top 8, bottom 392, scrolls inside |

Your 150/592 and 8/392 are what I got to the pixel. On `main` it was worse for me than your −7px (the picker landed *below* the viewport rather than above), which only strengthens the case.

Checked the one thing that would make the clamp lie — `pick.offsetHeight` is read *after* `document.body.appendChild(pick)` (lines 1765→1766), so it's the real box, not zero. Merges cleanly with today's `main`, builds, gates pass.

Ready from the slides side; it's waiting on a maintainer clearance, which is where every external PR waits here.

PR #450, opened 2026-09-12
Make it possible to theme code snippets.

tikurahul:
* Introduces the 0-tier palette for code.

**What & why**
Makes it possible to theme code snippets.

**How I verified it**
- `npm run-script build:single` and verified the changes to the model, as well as rendering.

**Checklist**
- [ ] Ran `node scripts/test-sync.ts` if I touched `slides/src/sync/`
- [x] Document format changes are additive and backward-compatible


nyblnet:
Verified both ways in a built shell, real Chrome:

- With `theme.codePalette = { c:'#ff0000', s:'#00ff00', n:'#0000ff', k:'#ff00ff' }`, the rendered tokens take exactly those colours — comment `rgb(255,0,0)`, string `rgb(0,255,0)`, number `rgb(0,0,255)`, keyword `rgb(255,0,255)`.
- With no `codePalette`, every token is byte-identical to today's defaults — `#6b7f8f`, `#c98a3e`, `#b0688f`, `#5b8def`. So existing decks don't move.

Merges cleanly with `main`, `tsc`, `build:single`, shell gate, `test-theme` 38/38, `test-tokenize` and `test-codediff` all pass. Additive: an older shell ignores the key and paints defaults, which is the right degradation. And `theme.chartPalette` already exists, so this is the natural sibling — the discussion in #446 was right that the theme is where it belongs.

Three small things, none blocking:

1. `doc.theme?.codePalette as any` — the type is declared right there in `model.ts`; reading it as the declared shape costs nothing and keeps the compiler on your side.
2. The palette can't set the **base text** colour: `DEFAULT_CODE_COLORS` falls through to `'x'` for a token with no scope, and `codePalette` has no `x`. If that's deliberate (base text follows `el.color`), a one-line comment saying so would stop the next reader wondering.
3. The defaults now live in two places — `code.ts` and the starter deck — with the same eight values. If they ever drift, the starter deck shows one thing and a fresh deck another. Having the starter deck spread the constant (or leaving the starter deck alone and letting it inherit) would keep one source of truth. Your call; declaring it on the starter deck is a nice way to show authors the knob exists.

Ready from the slides side.

PR #486, opened 2026-09-15
slides: compact input, round two — text fits itself, markdown accepted, a load reports what it dropped

nyblnet:
Stacked on #484 (`slides-compact-doc`); retargets to `main` once #484 merges. Compact-input only — the saved file is unchanged (save.ts/model.ts untouched, no model keys added, `modelkeys` unchanged).

**Text sizes itself.** In a compact document a text element may omit `h` (or write `"h": "auto"`). `compact.ts` (pure, node-importable) writes a provisional one-line height and lists the element; `compactload.ts` (browser) measures it with `measureElement` — the exact computation behind the panel's *Fit height to text* — on the deck's real fonts and writes the number before the document reaches the store, so undo never sees the provisional frame. If a deck font is still downloading at load, the report's `refit` names the elements and `loadDoc` fits them again after `document.fonts.ready` as one undo step. Tables are never fitted (their height follows rows).

**Markdown accepted.** `md` on a text element converts through `editor/markdown.ts`'s `markdownToHtml` — the same function a pasted plain-text block goes through — so bold, italic, code, strike, bullets, indented sub-bullets and `[caption](url)` come out as the editor would make them. `html` wins when both are present; `md` never reaches the document.

**A load says what it dropped.** The gate (`untrusted.ts`) gains `withDropReport`: when a collector is attached, every discarded key or value is recorded with a JSON-pointer path and a one-line reason (`/slides/0/elements/2/fontSze — unknown key for a text element`; `invalid value`; `unknown element type "sparkle" — element dropped`; `missing required … — object dropped`). Off by default: `note()` is one null check and the path trail is never touched, so pastes and remote ops cost what they did. `window.bento.loadDoc(json)` now returns `{ ok: true, compact, dropped, expanded, fitted, findings, refit }` (still `false` on non-documents, so `if (loadDoc(j))` reads as before); *Replace from JSON…* shows "Loaded: N fields dropped, M warnings — see console" and logs the full report.

**Agent-deck corpus.** `scripts/fixtures/agent-decks/` — three compact decks written the way agents write them (title + bullets with `md`; three cards from a helper's nested arrays with absolute geometry; a notes deck using `md`, omitted `h`, `"auto"` and nested arrays). `scripts/test-slides-agent-decks.ts` (esbuild-bundled, in CI) loads each through the real loader and asserts zero dropped keys, zero `validate()` errors, ids present and unique, no `md` left, provisional heights where `h` was omitted; and a deliberately wrong deck reports `bogus`, `onclick`, a bad `fontSize` and an unknown element type at their paths. 38 checks; negative control: silencing the unknown-key report fails 2. `test-slides-compact.ts` grows to 61 (md/html precedence, provisional height arithmetic, stats, report shape by source). `test-ci-registered` 339.

**Measured in Chrome** (built shell, real `loadDoc`): a 6-line `md` text at w 600, fontSize 24, no `h` → html `<b>Six lines</b> … • first bullet<br>• second bullet<br>  ◦ a sub-bullet …`, `h` = 180, `window.bento.measure()` on the same element = 180 (**delta 0**); an `"auto"` one-liner → 30 = measured 30; report: `dropped` = `/slides/0/elements/0/bogus` and `/onclick` ("unknown key for a text element") and `/slides/0/elements/2/type` (unknown element type, element dropped — 2 elements survive), `expanded` 24, `fitted` 1, `refit` 0 (fonts loaded), findings 0/0/0; `serialize()` emits the full document with no flag; `loadDoc('not json')` → `false`. *Replace from JSON…* through the Save ▾ menu: toast "Loaded: 2 fields dropped, 0 warnings — see console".

Strings: 1 × 30 catalogs (coverage 100%). AGENTS.md: the compact section gains the three additions and the load → read → fix → reload loop (repo file, stated). CHANGELOG entry under the #484 entry. `git diff origin/main -- CLAUDE.md docs/` empty.

**Added after review.** (1) The `size`/`theme`/`title`/`format`/`version` guard the bento-check build surfaced (a compact deck without `size` passed `parseDoc` and threw in `rebuildSidebar`) lives in **#484** (`slides-compact-doc` 8935507, six rig cases) and is merged into this branch (da99d57). Measured in Chrome: a compact deck with no `size`/`theme`/`title` loads, `size` = 1280×720, two sidebar thumbnails render, zero console errors; `size: null, theme: null` likewise. (2) **Gate parity rig** `scripts/test-gate-parity.ts` (in CI after `git fetch origin main`): main's `untrusted.ts` is taken from git at rig time and bundled beside the branch's; every checked-in deck (starter, 2 gallery fixtures, 3 agent decks expanded) plus a hostile corpus in the shapes the existing gate rigs cover — **50 slide fixtures, 5,514 elements compared individually** — goes through both with the collector OFF and the JSON outputs are asserted byte-identical; with the collector ON the same hostile slide reports 54 drops. Negative control: removing the required-keys check from the branch fails it 8 ways. For security's review, the collector hunk is `withDropReport`/`withPathSegment`/`note`/`within` in `untrusted.ts`: `note()` returns on a null collector before touching anything; `within()` runs `fn` directly when off.


nyblnet:
Superseded by #490: the same change rebuilt on the #484 squash without rewriting history.

PR #506, opened 2026-09-17
slides: pasting keeps its formatting; a long bullet wraps under its text (#503, #502)

nyblnet:
Two of Hermholtz's three asks from this morning, both about typing in text boxes (the third, #501, is #505 and this branch builds on it).

**#503 — pasting keeps the formatting.** Both edit-time paste handlers (text box, table cell) read only `text/plain` and rebuilt from markdown, so copying bold/italic/a list from one box and pasting into another lost it all. They now prefer the clipboard's `text/html` flavour through a new `editor/paste.ts` `clipboardToHtml`, which runs it through the ONE sanitizer (`render.ts sanitizeHtml`) after handling Chrome's clipboard quirks: the leading `<meta charset>`, `<style>`, the `<span style="…">` font soup, and `<b style="font-weight:normal">` — Chrome's way of writing "not bold" inside a bold run, which attribute-stripping alone would turn into a real bold; it is unwrapped instead. Plain text remains the fallback and keeps its markdown conversion; an html flavour that is nothing but junk hands over to it. `canvas.ts` never reads `text/html` directly (rig-asserted). The document-level ⌘C handler does not intercept while editing (measured: zero `writeText` calls on ⌘C in a box).

**#502 — a long bullet wraps under its text.** Typed `- ` stays a glyph while typing (markdown.ts explains why: converting the line under the caret moves the caret and captures Enter). At COMMIT, `editor/bullets.ts` `bulletsToLists` — a pure string function — turns runs of glyph lines into real `<ul><li>` (an indented `◦` nests under the item before it), so the renderer's existing list CSS gives the hanging indent everywhere (canvas, thumbnails, present, print). Lines are what `<br>` and contentEditable's `<div>` blocks separate; a box with no typed bullet is returned byte-identical; lists the bar made, a glyph mid-line, and a glyph typed inside an existing `<li>` are untouched; idempotent. Table cells are not converted (no list styling in a cell). Editing a box that already holds a list is native contentEditable: Enter inside an `<li>` makes the next item (measured).

**Measured in Chrome (built shell)**
- Type `- ` + a 70-character line + Enter + `  - a sub point`, commit → model `<ul><li>…<ul><li>a sub point</li></ul></li></ul>`; the wrapped first item's three line boxes all start at x = 300 while the box's text edge is at 267 (33 px hanging indent); the sidebar thumbnail carries the list.
- Paste Chrome's own clipboard html (`<meta charset>` + styled span + `<b>` + `<i>` + `<b style="font-weight:normal">plain</b>` + `<ul>`) → `<b>Bold</b> and <i>italic</i> plain<ul><li>one</li><li>two</li></ul>`.
- Paste plain `**strong** and *em*\n- item` → `<b>strong</b> and <i>em</i><ul><li>item</li></ul>`.
- Bar-made list, Enter after the item, type → `<ul><li>from the bar</li><li>second</li></ul>`.

**Security rig.** `test-sanitize` gains section 3d: a `DataTransfer` carrying the hostile corpus as `text/html` (onclick, onmouseover, `<script>`, javascript: href, a nested-unknown-tag case from #467, fixed-position style, ``, a beacon url) goes through `clipboardToHtml` and `insertHTML` into a live contentEditable; asserts formatting survives, "not bold" is unwrapped, no handler/script/style/img/meta survives, only the web link keeps its href with no other attribute, clicking and hovering every pasted node runs nothing, and the two hand-over cases. 154/154 in Chrome; negative control (sanitizer removed from the helper) fails five ways with `pwned 51,52,56`.

**Rigs**: `scripts/test-slides-bullets.ts` 27/27 (in CI), test-sanitize 154/154, links 33/33, ci-registered 355, typecheck, build:single 650 KB, shell gate. CHANGELOG two lines. No new strings, no docs/CLAUDE.md edits.

Not done: a paste from a non-Chrome source (Word, Safari) was not measured — the sanitizer decides the same way regardless of source; the Chrome quirks handled are the ones Chrome writes.


nyblnet:
Follow-up commit `bce682b8`, from review: `clipboardToHtml` now parses the clipboard's html with `DOMParser` into a document of its own, does the tidy-up (meta/style removal, the not-bold `<b>` unwrap) there, and hands that document's serialised body string to `sanitizeHtml`. The earlier version cloned the parsed nodes into a `div` created in the live document before sanitizing; a node adopted into the live document begins loading what it names and fires its load/error handlers, so the paste's markup could act before the sanitizer had seen it. Nothing in `paste.ts` creates, appends or clones into the page now (source-pinned in `test-slides-bullets`).

Rig: `test-sanitize` 3d pastes markup carrying img (with srcset), picture/source, video poster and src, audio, input type=image, noscript img, iframe and object — all addressed to the rig's own server — with load and error handlers; a fence request precedes the settle; the node side asserts the server received none of them and no handler ran. Against the previous helper: 7 requests received and handlers 61/62/63 ran (red); on this commit: 0 and none (green). 158/158.

PR #511, opened 2026-09-17
slides: a refused deck snapshot says the deck is too large; the Share panel counts pictures still uploading

nyblnet:
Slides side of #509, stacked on `kernel-snap-assets` (retargets to main once #509 merges).

- `syncNoticeText`: a `too-large` notice with `snapshot: true` (kernel's field for a refused whole-deck snapshot, which matches no op of ours) is worded "This deck is too large to share live in one piece. Your changes are saved in your copy, but a collaborator joining now may not receive the whole deck." — before the `media` branch; the change and image wordings are unchanged. After #509 a snapshot carries no inline pictures, so this now means the deck's own text/tables/small assets exceed the relay's frame ceiling (a comment says so).
- The toast on a snapshot refusal appends "N pictures are still uploading; they will follow." when `session.pendingBlobUploads() > 0` — informational, stated in the code.
- The Share popover gains a "N pictures still uploading…" line: `pendingBlobUploads()` polled once a second only while the popover is open (the interval stops itself when the wrap closes; a rebuild clears the previous poll), hidden at 0, singular/plural via the existing key pattern, only while sharing is on.
- Rig `scripts/test-slides-sync-notice.ts` 21/21, registered in CI (`snapshot` before `media`; the kernel type is used, not widened; the poll's open-only lifecycle; every new string in all 8 core catalogs); 5 strings × 30 catalogs, coverage 100%; ci-registered 359; typecheck; build 653 KB.
- Chrome: the built shell boots; the new strings resolve through the compiled catalogs (`2 pictures still uploading…` / de `2 Bilder werden noch hochgeladen…`). The popover line and the toast could not be driven live: the session instance is not reachable from the page (`window.bento.sync` is a facade) and turning sharing on reaches the real relay — their behaviour is source-pinned by the rig instead.
- Changelog line.


nyblnet:
Superseded by the fresh branch slides-sync-notice-wording-2 onto main e6a5389b (no retarget, by the team rule).

PR #519, opened 2026-09-19
kernel: remember the file handle across reopens (one prompt, then silent, for Downloads/Documents)

nyblnet:
Maintainer decision (2026-09-19, "let's do 1 and 2 now"), relayed by home-webext. This is item 1. Spec: `working/team/handoffs/persist-file-handle.md`.

## Why
Chrome's File System Access blocklist refuses `~/Documents`, `~/Downloads`, the Desktop and the home folder as **directory** grants (to every page and extension). So a deck sitting directly in one of them can never be covered by a bento/home folder grant, and today every ⌘S after a reopen re-runs the full save picker. **FILE** handles are not blocked and are structured-cloneable — so we remember one.

## What (kernel/src/save.ts)
- **Persist** the `FileSystemFileHandle` a save produced in IndexedDB (`bento-handles`, keyed by `docId`), on both the picked and the adopted (dropped-file) save.
- **Reconnect** on the first save after a reopen — `reconnectHandle(doc)`: `queryPermission` then, if needed, `requestPermission({mode:'readwrite'})` **inside the save gesture**. Chrome 122+ shows a small "Allow on every visit" prompt instead of a picker. Granted → reuse; denied or `NotFoundError` (moved/deleted) → drop the entry, fall back to the picker.
- **Before serialize:** `saveFile` reconnects before `serializeAuto`, so the transient user activation is still fresh when the prompt fires (serialize can take ~1s on a big deck — thanks @home-webext for the activation note).

## Guards
- **name-match** — the stored handle's `name` must equal the file on screen (`docId` + name; `isSameEntry` needs a second handle a bare `file://` reopen lacks). Mismatch → picker, entry kept.
- **real handle only** — `instanceof FileSystemFileHandle`. A host polyfill (home/ios, home/webext) hands back its own handle, which is not one, so it is never stored and **the bridge path is untouched**. A granted-folder deck keeps going through the extension exactly as before.
- **readonly players** never save, so their handle is never stored. Encrypted decks are fine (a handle is not plaintext).

## Rig — `scripts/test-file-handle.ts` (18 checks, mutation-verified)
Drives `reconnectHandle` against a fake IndexedDB + FileSystemFileHandle: granted → reconnect, no prompt; prompt→granted → reconnect; denied → picker + entry dropped; NotFoundError → picker + entry dropped; name mismatch → picker, entry kept; host polyfill → skipped; nothing stored → picker. Plus a source check that `saveFile` reconnects before it serializes. Mutation-tested: removing the name-match / drop-on-denied / instanceof guard each goes red on the matching check.

## Verification
test-file-handle 18/18 · test-savepurpose 16/16 · test-export-secrets 74/74 · test-ci-registered 359/359 · kernel `tsc` clean.

@home-webext to run it against the extension's paths: a deck in a granted folder (bridge, no prompt) vs one in ~/Downloads (this path: one prompt, then silent).

## Known interaction (home-webext review)

A deck saved **without** the extension (so a real handle was stored) whose folder is **later** granted takes this reconnect path first — one "Allow on every visit" prompt — rather than the bridge's silent claim, because reconnect runs before the picker and cannot know the host would claim. Rare (only decks saved before the grant existed), self-healing after that one prompt, and the alternative ordering (picker/bridge first) would lose the stored handle for the ~/Downloads case the whole feature exists for. The ordering is intentional. Bridge path otherwise verified identical to today (fileHandle null → reconnect finds nothing stored, since the polyfill handle is never persisted → pickHandle → the polyfill claims and answers with no dialog → persist skips it again; rig case 6 pins the skip).


nyblnet:
Superseded by #533 — cherry-picked onto current main (this branch was based on 1.2.1 and went stale), CHANGELOG resolved, and the shared-origin gate added (persistence + reconnect OFF on file:// where storage is shared; on for real origins). Continue review on #533.

PR #521, opened 2026-09-19
packs: the URL override reads from the host

nyblnet:
Off #520's branch (`kernel-autosave-trim`), one commit; rides with it.

`slides/src/packs.ts`: the `bento-packs-url` dev override is read only where storage is this document's own (`sharedStorageOrigin()` false); on a shared origin the configured default channel is used. `scripts/test-packs.ts`: three cases — ignored on the shared default (node has no `location`, the fail-safe reading), honoured on a real web origin, ignored on `file://`. 25/25. No changelog or decisions line.


nyblnet:
Superseded by the fresh branch packs-url-host-2 onto main 870f34ec (no retarget, by the team rule).

